PR #6, opened 2018-05-17
Fix .Aggregate exception if replay has no mods

smoogipoo:
TIL .Aggregate throws an exception if the collection is empty!

### DIFF
--- a/PerformanceCalculator/Performance/PerformanceProcessor.cs
+++ b/PerformanceCalculator/Performance/PerformanceProcessor.cs
@@ -40,7 +40,8 @@ namespace PerformanceCalculator.Performance
 
                 command.Console.WriteLine(f);
                 command.Console.WriteLine($"{"Player".PadRight(15)}: {score.User.Username}");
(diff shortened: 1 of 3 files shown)


PerformanceCalculator/Performance/PerformanceProcessor.cs
```diff
@@ -40,7 +40,8 @@ public void Execute()
 
                 command.Console.WriteLine(f);
                 command.Console.WriteLine($"{"Player".PadRight(15)}: {score.User.Username}");
-                command.Console.WriteLine($"{"Mods".PadRight(15)}: {score.Mods.Select(m => m.ShortenedName).Aggregate((c, n) => $"{c}, {n}")}");
+                if (score.Mods.Length > 0)
+                    command.Console.WriteLine($"{"Mods".PadRight(15)}: {score.Mods.Select(m => m.ShortenedName).Aggregate((c, n) => $"{c}, {n}")}");
```
peppy: `else "Mods:          None"`

PR #12, opened 2018-12-28
Add simulate command in PerformanceCalculator simulating performance

qwewqa:
Still needs some testing and cleanup

Adds a simulate subcommand to PerformanceCalculator which calculates the performance (and subattributes) a play with the given details would have on a beatmap.

### DIFF
--- a/PerformanceCalculator/README.md
+++ b/PerformanceCalculator/README.md
@@ -106,11 +106,11 @@ Run 'simulate [command] --help' for more information about a command.
 Computes the performance of a simulated play on a beatmap. The provided output includes raw performance attributes and pp value.
 
 
-#### Osu
+#### osu!
 ```
 > dotnet PerformanceCalculator.dll simulate osu --help
 
-Computes the performance (pp) of a simulated osu play.
+Computes the performance (pp) of a simulated osu! play.
 
 Usage: dotnet PerformanceCalculator.dll simulate osu [arguments] [options]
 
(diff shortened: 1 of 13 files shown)


PerformanceCalculator/README.md
```diff
@@ -81,4 +82,83 @@ Aim            : 123.614719845539
...
+  osu           Computes the performance (pp) of a simulated osu play.
+  taiko         Computes the performance (pp) of a simulated taiko play.
+
+Run 'simulate [command] --help' for more information about a command.
+
+```
+Computes the performance of a simulated play on a beatmap. The provided output includes raw performance attributes and pp value.
+
+
+#### Osu
```
HoLLy-HaCKeR: Should probably be `osu!`.

PerformanceCalculator/Simulate/Osu/OsuSimulateProcessor.cs
```diff
@@ -0,0 +1,119 @@
...
+            writeAttribute("Misses", statistics[HitResult.Miss].ToString(CultureInfo.InvariantCulture));
+
+            writeAttribute("Mods", mods.Length > 0
+                ? mods.Select(m => m.Acronym).Aggregate((c, n) => $"{c}, {n}")
+                : "None");
+
+            foreach (var kvp in categoryAttribs)
+                writeAttribute(kvp.Key, kvp.Value.ToString(CultureInfo.InvariantCulture));
+
+            writeAttribute("pp", pp.ToString(CultureInfo.InvariantCulture));
```
smoogipoo: Can we remove this duplication by moving most of this code and that of `getMods()` into a base class (`SimulateProcessor`) that derives from `IProcessor`?

It seems like these classes should only contain `generateHitResults()` (base class should provide as abstract method).

qwewqa: Yes

PR #14, opened 2019-01-24
Profile calculator

peppy:
Replaces #13.

Still requires some fixing.

PR #15, opened 2019-01-24
Combine commands and processors

peppy:
- [x] Depends on #14.

There was no reason to separate out the commands from their logic.

Makes it easier to access paramters. Also exposes formatting via `OutputDocument`.


peppy:
Conflict resolved

PR #26, opened 2019-02-19
Tiny QOL change for ProfileCommand

Victor4X:
Local PP now also displays the difference to Live PP.  
This makes it easier to see improvements at a glance, and was a QOL change I made when I was helping in the reddit PP Calculation Thread: https://www.reddit.com/r/osugame/comments/al27eu/pp_calculation_thread/   

Example:
image

### DIFF
--- a/PerformanceCalculator/Profile/ProfileCommand.cs
+++ b/PerformanceCalculator/Profile/ProfileCommand.cs
@@ -95,7 +95,6 @@ namespace PerformanceCalculator.Profile
             int index = 0;
             double totalLocalPP = localOrdered.Sum(play => Math.Pow(0.95, index++) * play.LocalPP);
             double totalLivePP = userData.pp_raw;
-            double totalDiffPP = totalLocalPP - totalLivePP;
 
             index = 0;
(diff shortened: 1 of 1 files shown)


peppy:
Can you change the formatting to
`Local PP: 5562,0 (+219,5)`?

I think the format should be `$"{totalDiffPP:+F1;-F1;-}"` (please check this is correct).

tybug:
Should rather be `{totalDiffPP:+#.#;-#.#;-0}` for single decimal precision and a `-0` for 0 difference

peppy:
It should be a `-` with no difference, or `+0`, but not `-0`.

tybug:
Ah okay, I thought `Local PP: 5562,6 (- Difference)` might look a bit strange with no number next to the sign

Victor4X:
Hey! I've removed the word "Difference" and updated the number formatting to include the sign.
With this implementation it rounds to a single decimal, and looks like this:

If positive:
if positive

If negative:
if negative

If zero:
if zero

And here is how it would look with double decimal precision:
if positive 2 digits

PR #29, opened 2019-03-20
Enable tiered compilation to calculate faster.

bltsheep:
Tiered compilation can make calculation faster, especially when there are only a few beatmaps. See https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-2-1#jit-compiler-improvements

Here are the comparisons.
image


smoogipoo:
We've had issues in the past with tiered compilation, some of which occurred even in netcore3.0. We'll enable tiered compilation after netcore3.0 is released (it'll be made default) and we can safely assure the JIT is stable.

HoLLy-HaCKeR:
Off topic, I also tried this before on lazer, but didn't notice any performance changes.

PR #52, opened 2020-09-25
Add --json arg for simulate command and refactor to use OutputDocument method

Syriiin:
- Add `--json` arg for simulate command
- Refactor simulate command to use `OutputDocument` rather than `Console.WriteLine`

The json could definitely be formatted better (not using floats for all numbers, mods as bitmask, etc...) but that's a side effect of trying to keep it as close to the regular output as possible.
The purpose of this is mainly to make it easier for other code to consume the output without unnecessary parsing.

Example
```sh
$ dotnet run -- simulate osu "D:\beatmaps\546514" --json
{
  "Beatmap": "Printemps - Eien Friends (Sakaue Nachi) [Friends]",
  "Accuracy": 72.90779547992979,
  "Combo": 1599.0,
  "Great": 1230.0,
  "Good": 0.0,
  "Meh": 0.0,
  "Miss": 0.0,
  "Mods": "None",
  "Aim": 84.74871684189024,
  "Speed": 42.877486554927245,
  "OD": 7.833333333333333,
  "AR": 9.0,
  "Max Combo": 1599.0,
  "pp": 203.99553541963314
}
```

### DIFF
--- a/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
@@ -99,15 +99,19 @@ namespace PerformanceCalculator.Simulate
             return hits / total;
         }
 
-        protected override void WritePlayInfo(ScoreInfo scoreInfo, IBeatmap beatmap)
+        protected override string GetPlayInfo(ScoreInfo scoreInfo, IBeatmap beatmap)
         {
-            WriteAttribute("ApproachRate", FormattableString.Invariant($"{beatmap.BeatmapInfo.BaseDifficulty.ApproachRate}"));
-            WriteAttribute("MaxCombo", FormattableString.Invariant($"{scoreInfo.MaxCombo}"));
+            var playInfo = new List<string>();
+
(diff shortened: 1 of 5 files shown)


peppy:
Seems to work fine, but this is a pretty non-OO approach to this, but not sure how much we want to scrutinise this project as opposed to refactoring later if required. @smoogipoo thoughts?

Syriiin:
What kind of refactors would you want to see?
I was intending to refactor all the commands to allow json output, but I wasn't sure how to best do that while preserving the procedural nature of `Execute`.
Perhaps returning an object representation of the output from `Execute`, and formatting that into either a `JObject`, or a `Document` in `ProcessorCommand`?
Would just want to be wary that this kind of refactor wouldn't change the regular structure of the output, for the sake of backwards compat (since there are likely already projects that rely on parsing that exact format).

smoogipoo:
@peppy Yeah I don't think it's worth spending too much time refactoring this project right now.

smoogipoo (approved):

PR #54, opened 2020-11-10
Fix osu!catch performance calc in profile command.

ilsubyeega:
¯\_(ツ)_/¯
[CatchPerformanceCalculator.cs](https://github.com/ppy/osu/blob/dd7c3dbc5be06743d00ef3a95b450ef23819194b/osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs) additionally requires LargeTickHit, SmallTickHit, SmallTickMiss
image

tested this pr's branch with osu!, osu!taiko, osu!catch but only osu!mania is incorrect [(mentioned this here)](https://twitter.com/ilsubyeega/status/1325825555447513089)

### DIFF
--- a/PerformanceCalculator/Profile/ProfileCommand.cs
+++ b/PerformanceCalculator/Profile/ProfileCommand.cs
@@ -72,11 +72,14 @@ namespace PerformanceCalculator.Profile
                     Mods = mods,
                     Statistics = new Dictionary<HitResult, int>
                     {
-                        { HitResult.Perfect, (int)play.countgeki },
+                        { HitResult.Perfect, (Ruleset == 2) ? (int)play.count300 : (int)play.countgeki },
                         { HitResult.Great, (int)play.count300 },
                         { HitResult.Good, (int)play.count100 },
                         { HitResult.Ok, (int)play.countkatu },
                         { HitResult.Meh, (int)play.count50 },
+                        { HitResult.LargeTickHit, (int)play.count100 },
+                        { HitResult.SmallTickHit, (int)play.count50 },
+                        { HitResult.SmallTickMiss, (int)play.countkatu },
                         { HitResult.Miss, (int)play.countmiss }
                     }
                 });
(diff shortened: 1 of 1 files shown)


PerformanceCalculator/Profile/ProfileCommand.cs
```diff
@@ -72,11 +72,14 @@ public override void Execute()
                     Mods = mods,
                     Statistics = new Dictionary<HitResult, int>
                     {
-                        { HitResult.Perfect, (int)play.countgeki },
+                        { HitResult.Perfect, (Ruleset == 2) ? (int)play.count300 : (int)play.countgeki },
```
smoogipoo, line 75: The performance calculator is wrong. It should be changed to use `HitResult.Great`.

smoogipoo (changes requested):
Rest is fine

PerformanceCalculator/Profile/ProfileCommand.cs
ilsubyeega: so should we update osu submodule?

ilsubyeega:
I'll just match this PR to the [latest CatchPerformanceCalculator](https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs).

PerformanceCalculator/Profile/ProfileCommand.cs
smoogipoo: Yeah. I was holding off updating it because we'd need to go through the simulate command and see if it's generating the correct statistics.

PerformanceCalculator/Profile/ProfileCommand.cs
```diff
@@ -77,6 +77,9 @@ public override void Execute()
                         { HitResult.Good, (int)play.count100 },
                         { HitResult.Ok, (int)play.countkatu },
                         { HitResult.Meh, (int)play.count50 },
+                        { HitResult.LargeTickHit, (int)play.count100 },
+                        { HitResult.SmallTickHit, (int)play.count50 },
+                        { HitResult.SmallTickMiss, (int)play.countkatu },
```
smoogipoo, line 82: I still don't really like this, but have a solution for you. Extract this `new ScoreInfo` into a variable and use the `ScoreInfoExtensions` to set the properties. So like `scoreInfo.SetCount100((int)play.count100))`.

smoogipoo (changes requested):

ilsubyeega:
Suprised that it still works

smoogipoo (approved):

PR #56, opened 2021-01-18
Remove osu! submodule

smoogipoo:


stanriders:
This is gonna be really inconvenient for anyone working on difficalc because you'd either have to manually readd osu into the project or export nuget packages every time you compile it

bdach:
From a maintenance standpoint, with nuget packages dependabot can be leveraged to submit updates, in comparison to the current mode of "this gets updated whenever somebody remembers to". Which doesn't work, based on cases like the one #55 was fixing (`simulate` giving wrong results) which apparently people knew for months and never actually bothered to report/submit a fix for.

Local project checkouts take like five seconds - I would know since I do it all the time with framework. [We even have scripts for that ready.](https://github.com/ppy/osu-framework/wiki/Testing-local-framework-checkout-with-other-projects)

peppy:
Yep, we run that linked script many times a day; it's low overhead and a necessary detail when creating nuget packages across linked projects. It should take a few minutes to make it work here (hopefully someone can do this and we'll post/link on the local wiki).

peppy (approved):

PR #69, opened 2021-03-21
add UseLocalOsu.ps1

joseph-ireland:
Add a script to more easily use a local copy of osu rather than a version from nuget. 

It's derived from the script linked here: https://github.com/ppy/osu-tools/pull/56

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,16 @@ Tools for [osu!](https://osu.ppy.sh).
 
 # Getting Started
 
+## I just want to run it
 - Clone the repository including submodules (`git clone --recurse-submodules https://github.com/ppy/osu-tools`)
 - Navigate to each tool's directory and follow the instructions listed in the tool's README.
     - [PerformanceCalculator](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/README.md) - A tool for calculating the difficulty of beatmaps and the performance of replays.
+
+## I want to make changes
+Most relevant code is in the main [ppy/osu](htts://github.com/ppy/osu) repository. To make any meaningful changes you will likely need to edit that as well.
+
+- Clone all relevant repos into the same directory (`git clone https://github.com/ppy/osu-tools`; `git clone --recurse-submodules https://github.com/ppy/osu`)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -11,10 +11,17 @@ Tools for [osu!](https://osu.ppy.sh).
...
 
+## I just want to run it
 - Clone the repository including submodules (`git clone --recurse-submodules https://github.com/ppy/osu-tools`)
 - Navigate to each tool's directory and follow the instructions listed in the tool's README.
     - [PerformanceCalculator](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/README.md) - A tool for calculating the difficulty of beatmaps and the performance of replays.
 
+## I want to make changes
+Most relevant code is in the main [ppy/osu](htts://github.com/ppy/osu) repository. To make any meaningful changes you will likely need to edit that as well.
+
+- Clone all relevant repos into the same directory (`git clone https://github.com/ppy/osu-tools`; `git clone --recurse-submodules https://github.com/ppy/osu`)
```
bdach, line 22: The `--recurse-submodules` seems superfluous here...? Neither repository has submodules anymore as far as I know, just nuget packages.

```suggestion
- Clone all relevant repos into the same directory (`git clone https://github.com/ppy/osu-tools`; `git clone https://github.com/ppy/osu`)
```

bdach (commented):
Might also include the bash script while we're at it.

joseph-ireland:
I wouldn't be confident pushing up a bash version since I'm not building on linux, I agree that'd be nice though

peppy:
Usually we just keep these kinds of scripts living on the wiki, but nothing against having this present if it gives it more visibility.

joseph-ireland:
Wiki is fine for osu framework, you'd expect a level of proficiency before people attempt to contribute there. But people looking at osu-tools are likely less experienced, and this action is probably one of the first things most people want to do.

I added a shell script. It's only tested with git bash though since I don't have dotnet installed in any linux environments.

smoogipoo:
Lgtm. Thanks!

smoogipoo (approved):

Issue #25, opened 2019-02-13
osu!mania mod multipliers not applied to score


wanjiaXG:
[picture](https://osu.wanjiaxg.com/mania_bug.png)
EZ NF HT，The results of these mods are wrong. like this picture, scores need to be recalculated.
[fix](https://osu.wanjiaxg.com/fix_mania_score.png)
This is my change(Is bad Code)

PR #85, opened 2021-05-31
Fix incorrect mania simulation score with HT/EZ/NF

smoogipoo:
Fixes https://github.com/ppy/osu-tools/issues/25

### DIFF
--- a/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
@@ -1,14 +1,17 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


peppy:
Can you sum up in a couple of sentences why we're only applying for diff reduction and why it matters? These scores aren't matching stable or anything in the first place right? I feel like I might be missing something.

smoogipoo:
> we're only applying for diff reduction

Stable doesn't add multipliers for difficulty increasing mods.

> why it matters

Because mania's performance calculator considers the score achieved:
https://github.com/ppy/osu/blob/bd7814b487797b8f10130eb5c640b3a725d39b7e/osu.Game.Rulesets.Mania/Difficulty/ManiaPerformanceCalculator.cs#L85-L107

> These scores aren't matching stable or anything in the first place right?

They are. You can construct complete score simulations using these commands, and that's the point of them.

peppy:
Right... didn't realise the performance calculations were so closely tied to total score in mania.

peppy (approved):

PR #87, opened 2021-05-31
Use osu! API v2 for profile command

### LINKED ISSUE #84: Use osu-web APIv2
The profile command currently uses APIv1. This tool should eventually be converted to APIv2 (see: https://docs.ppy.sh) to allow calculation of a larger set of the user's top scores.

stanriders:
Resolves #84 

Seems like v2's user best scores can return 51 scores max, which is less than v1's 100, might wanna create an issue on osu-web for that


smoogipoo:
V2 is paginated, so you can apply limit and offset. That should allow much more than 51 scores no?

stanriders:
> V2 is paginated, so you can apply limit and offset. That should allow much more than 51 scores no?

We can run 2 queries sure, but wouldn't it be simpler to just allow accessing all 100 on web side? It's the maximum API allows anyway, `offset=100` returns nothing

smoogipoo:
Hmm, I didn’t expect that. Definitely something for the web team to look into.

smoogipoo:
Going to block this one until osu-web allows retrieval of more than 50 scores. Because that is, at the very minimum, not enough.

stanriders:
@smoogipoo this should be good to go now

PR #110, opened 2021-08-03
Fix several issues when running profile command on mania

smoogipoo:
Two fixes:
1. Fixed a crash because mania doesn't output a `Max Combo` property from its pp calculator.
2. Fixed running calculation using mods that aren't relevant for difficulty calculation (i.e. that aren't stored to the database). In this case it was the MR mod.

### DIFF
--- a/PerformanceCalculator/Profile/ProfileCommand.cs
+++ b/PerformanceCalculator/Profile/ProfileCommand.cs
@@ -94,7 +94,6 @@ namespace PerformanceCalculator.Profile
 
                 var categories = new Dictionary<string, double>();
                 var localPP = performanceCalculator.Calculate(categories);
-                var maxCombo = categories["Max Combo"];
                 var thisPlay = new UserPlayInfo
                 {
                     Beatmap = working.BeatmapInfo,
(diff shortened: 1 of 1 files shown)


peppy (approved):

PerformanceCalculator/Profile/ProfileCommand.cs
```diff
@@ -67,16 +67,20 @@ public override void Execute()
...
                 var scoreInfo = new ScoreInfo
                 {
                     Ruleset = ruleset.RulesetInfo,
                     TotalScore = play.score,
                     MaxCombo = play.maxcombo,
-                    Mods = mods,
+                    // Consider only the mods relevant to difficulty calculation in order to match live values.
+                    // For example, the osu!mania MR mod changes SR, but is not part of the mod combinations stored to the database.
+                    // Todo: This can probably be removed when osu!lazer is used for realtime pp+diffcalc.
+                    Mods = ruleset.ConvertFromLegacyMods((LegacyMods)(play.enabled_mods) & difficultyCalculationMods).ToArray(),
```
stanriders, line 83: fyi this breaks HD scores on stardard ruleset

bdach: Just for the benefit of everyone being able to understand the problem, this breaks any mods that affect performance points but *not* star difficulty for that matter, correct? And the "breakage" is the pp values being wrong?

smoogipoo: Right, makes sense. I’ll need to separate into two calculations, thanks for the report.

To answer the above: yes it’ll break PP only, SR will be correct.

PR #121, opened 2021-09-29
Add flashlight attribute and update packages

smoogipoo:


MBmasher:
Does `FlashlightRating` have a zero value when Flashlight mod isn't enabled?

smoogipoo:
No, is that an issue worth fixing?

MBmasher:
I'm actually not sure. I feel like it makes sense for it to be zero when it's not enabled, although it's a weird case.

peppy:
I would honestly hope we aren't storing it anywhere when flashlight is off? Seems like a huge waste of resources otherwise.

smoogipoo:
Originally it wasn't 0 in the difficulty calculator because it wasn't included into SR and so it could be pulled from the base mod combinations (combinations excluding FL).

But now that diffcalc runs over +FL combinations as well, this should be changed osu!-side.

smoogipoo:
This PR can go in as is in any case.

peppy:
So we do want to show "flashlight rating" as zero on all outputs from this tool?

smoogipoo:
I've changed it to not output.

peppy (approved):

PR #122, opened 2021-10-09
Make `difficulty` command download map when given a beatmap ID

stanriders:
Similar to how `simulate` and `profile` do that

### DIFF
--- a/PerformanceCalculator/Difficulty/DifficultyCommand.cs
+++ b/PerformanceCalculator/Difficulty/DifficultyCommand.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using Alba.CsConsoleFormat;
 using JetBrains.Annotations;
 using McMaster.Extensions.CommandLineUtils;
+using osu.Framework.IO.Network;
 using osu.Game.Beatmaps;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Catch.Difficulty;
@@ -24,8 +25,8 @@ namespace PerformanceCalculator.Difficulty
     {
         [UsedImplicitly]
         [Required, FileOrDirectoryExists]
-        [Argument(0, Name = "path", Description = "Required. A beatmap file (.osu), or a folder containing .osu files to compute the difficulty for.")]
-        public string Path { get; }
+        [Argument(0, Name = "path", Description = "Required. A beatmap file (.osu), beatmap ID, or a folder containing .osu files to compute the difficulty for.")]
+        public string Path { get; private set; }
 
         [UsedImplicitly]
         [Option(CommandOptionType.SingleOrNoValue, Template = "-r|--ruleset:<ruleset-id>", Description = "Optional. The ruleset to compute the beatmap difficulty for, if it's a convertible beatmap.\n"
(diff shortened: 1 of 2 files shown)


PerformanceCalculator/Difficulty/DifficultyCommand.cs
```diff
@@ -23,9 +24,9 @@ namespace PerformanceCalculator.Difficulty
     public class DifficultyCommand : ProcessorCommand
     {
         [UsedImplicitly]
-        [Required, FileOrDirectoryExists]
-        [Argument(0, Name = "path", Description = "Required. A beatmap file (.osu), or a folder containing .osu files to compute the difficulty for.")]
-        public string Path { get; }
+        [Required]
+        [Argument(0, Name = "path", Description = "Required. A beatmap file (.osu), beatmap ID, or a folder containing .osu files to compute the difficulty for.")]
+        public string Path { get; private set; }
```
peppy, line 29:
```suggestion
[UsedImplicitly]
        [Required]
        [Argument(0, Name = "beatmap", Description = "Required. Can be either a path to beatmap file (.osu) or beatmap ID.")]
        public string Beatmap { get; set; }
```

Use the same as used in other implementations. It can't be called path if it isn't always a path.

PerformanceCalculator/Difficulty/DifficultyCommand.cs
```diff
@@ -59,7 +60,34 @@ public override void Execute()
                 }
             }
             else
+            {
+                if (!Path.EndsWith(".osu"))
```
peppy, line 64: This logic *really* should be moved out to a common location rather than being copied and pasted in more and more locations.. either a base class or a static helper at very least.

peppy (changes requested):

PerformanceCalculator/Difficulty/DifficultyCommand.cs
smoogipoo: You completely changed the semantics of this. I want to keep it as a path.

stanriders: This command allows a whole directory to be processed, while others only take singular beatmaps

PerformanceCalculator/ProcessorWorkingBeatmap.cs
```diff
@@ -46,6 +48,30 @@ private static Beatmap readFromFile(string filename)
...
+            if (fileOrId.EndsWith(".osu"))
+            {
+                if (!File.Exists(fileOrId))
+                    throw new ArgumentException($"Beatmap file {fileOrId} does not exist.");
+
+                return new ProcessorWorkingBeatmap(fileOrId);
+            }
+
+            if (!int.TryParse(fileOrId, out var beatmapId))
+                throw new ArgumentException("Incorrect beatmap ID.");
```
peppy, line 62:
```suggestion
                throw new ArgumentException("Could not parse provided beatmap ID.");
```

minor but "incorrect" reads a bit weird when it's actually invalid.

PerformanceCalculator/Difficulty/DifficultyCommand.cs
peppy: Super messy. I'm just going to turn a blind eye.

PR #135, opened 2021-11-19
Update `{Calculate -> Populate}Accuracy` in line with game-side changes

bdach:
Companion piece to https://github.com/ppy/osu/pull/15657.

#134 also contains this change, so if that gets merged immediately, it will supersede this pull.

- [ ] Depends on game bump.


peppy:
Seems to already be addressed in https://github.com/ppy/osu-tools/pull/134/commits/b57c977edd2695b810cdabb4d70ad82ef4ab7047

bdach:
Oh, well spotted. Kind of got smuggled in that other diff...

I guess I'll mark as an alternative for that one for now or something? Dunno how to proceed best really.

peppy:
That one will likely be merged since I've already reviewed it, just waiting for a package bump. Wasn't marked as waiting on a package bump so easy to miss.

Will close this for now.

PR #145, opened 2022-02-22
Add support for lazer-first scores

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/16575

- Added for `difficulty` and `simulate` commands if you provide the `-l` or `--lazer` arguments. I think this is best for now since classic scores are the most common.
- Added for `performance` command if you provide a lazer replay.

### DIFF
--- a/PerformanceCalculator/Difficulty/DifficultyCommand.cs
+++ b/PerformanceCalculator/Difficulty/DifficultyCommand.cs
@@ -43,8 +43,8 @@ namespace PerformanceCalculator.Difficulty
         public bool OutputJson { get; }
 
         [UsedImplicitly]
-        [Option(Template = "-l|--lazer", Description = "Excludes the classic mod to compute lazer-first scores.")]
-        public bool Lazer { get; }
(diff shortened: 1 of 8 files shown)


bdach:
Heads up that the dependency is sorted here and it may be a good idea to get this one out before people start reporting that their max combo is wrong. I don't have any perms here so I can't do anything about it :(

PerformanceCalculator/Difficulty/DifficultyCommand.cs
```diff
@@ -42,6 +42,10 @@ public class DifficultyCommand : ProcessorCommand
         [Option(Template = "-j|--json", Description = "Output results as JSON.")]
         public bool OutputJson { get; }
 
+        [UsedImplicitly]
+        [Option(Template = "-l|--lazer", Description = "Excludes the classic mod to compute lazer-first scores.")]
+        public bool Lazer { get; }
```
peppy, line 47: @smoogipoo would you be against renaming these to `-nc` / `NoClassicMod` or something similar? Rationale is to avoid adding lazer in variables (we've avoided it quite well until now).

peppy (approved):

PR #147, opened 2022-03-14
Apply osu!-side PerformanceCalculator changes

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/17242


Natelytle:
For some reason the combo argument is being multiplied by the max combo of the map / 100, can you see if you can replicate it
image
image

peppy (approved):

PR #152, opened 2022-06-14
Add mod definition output command

peppy:
To be used for more automated web-side updates when mod definitions change.

Usage: `dotnet run -- mods`


smoogipoo (approved):

stanriders:
I'm a bit late, but I feel like this doesn't belong to the PerformanceCalculator, wouldn't it be more fitting to create a separate project for this? osu-tools is supposed to be a repository of *multiple* tools, not just the perfcalc

peppy:
I'd sooner consider renaming the project to just be `osu.Tools` to be honest.

I don't see anything gained from splitting up further.

PR #154, opened 2022-07-16
Use new score data format

stanriders:


peppy (approved):
Looks okay at a surface level

smoogipoo (approved):

PR #156, opened 2022-07-25
Fix up ProfileScreen + various cleanups

smoogipoo:
Fixes up a double increment in `ProfileScreen` and applies some refactoring to reduce code duplication, by using the base `PerformanceCalculator` project.

I've added a `LoginCommand` to the CLI in the process, which is better than having commands of e.g. `profile <user_id> <client_id> <client_secret>`. I constantly run into this when I want to debug/use the tool, so I think this will be a welcome QOL change.

We may want to make the `GameHost`s non-portable, so the same credentials can be shared between the CLI and GUI versions?


PerformanceCalculatorGUI/Program.cs
```diff
@@ -12,7 +12,7 @@ public static class Program
         [STAThread]
         public static void Main(string[] args)
         {
-            using DesktopGameHost host = Host.GetSuitableDesktopHost("PerformanceCalculatorGUI", new HostOptions
+            using DesktopGameHost host = Host.GetSuitableDesktopHost("PerformanceCalculator", new HostOptions
             {
                 PortableInstallation = true,
```
smoogipoo, line 17: There's no functional change as a result of this since it's a portable install. But as I mentioned in the OP, perhaps we should make this non-portable to share creds with CLI.

peppy: I mentioned in discord, but I'd also see things go in the direction of using `APIAccess` rather than reimplementing here, if possible.

peppy:
Needs unused file cleanup https://github.com/ppy/osu-tools/runs/7492441335?check_suite_focus=true

PerformanceCalculatorGUI/Program.cs
peppy: From that angle i'm not too sure about making the new `APIManager` class inside the non-GUI version. I guess if we're holding all projects in this repo to the same standard that's fine, dunno. Will go with it for now I guess.

peppy:
This is a breaking change for existing usages of the api right? Anyone updating to this will have to change their workflow to first login, and all existing commands using login via parameters will cease to work?

I'm a touch unsure about the change... I feel like the CLI version shouldn't be relying on a `GameHost` to keep things as simple as possible. Maybe just me.

smoogipoo:
Hmmm, alright, I'll PR just the necessary changes and leave the API stuff to another time. I don't want discussion about this, if it wasn't an immediate merge then better to close for me.

PR #164, opened 2022-09-09
Fix mod definition output handling `Description` incorrectly

peppy:
Bumps `osu.Game` and fixes handling of `LoclisableString`.


peppy:
Have tested that this has a nil diff with current `osu-web` json file.

smoogipoo (approved):

PR #166, opened 2022-10-13
Simulate osu!mania scores by accuracy and counts

kionell:
Since osu!mania now uses accuracy -> pp system instead of score -> pp, mania score simulation needs to be updated in both osu-tools and osu-tools GUI. However, i heard that you want to revert this change so idk if this pull request is still relevant or not.

### DIFF
--- a/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
@@ -53,7 +53,7 @@ namespace PerformanceCalculator.Simulate
         protected override int GetMaxCombo(IBeatmap beatmap) => beatmap.HitObjects.Count(h => h is Fruit)
                                                                 + beatmap.HitObjects.OfType<JuiceStream>().SelectMany(j => j.NestedHitObjects).Count(h => !(h is TinyDroplet));
 
-        protected override Dictionary<HitResult, int> GenerateHitResults(double accuracy, IBeatmap beatmap, int countMiss, int? countMeh, int? countGood)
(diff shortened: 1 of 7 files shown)


kionell:
Also, is there any reason why in [ScoreInfoExtensions](https://github.com/ppy/osu/blob/master/osu.Game/Scoring/Legacy/ScoreInfoExtensions.cs#L92) all 100's are Ok's, while everywhere else they are Good's?

smoogipoo:
> while everywhere else

Where do you mean?

S1n1st3rn3ss:
[OsuSimulateCommand](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/OsuSimulateCommand.cs#L47) as one of the cases

kionell:
> Where do you mean?

Oh, at first i thought that this is the case for osu!taiko simulate command and how osu-web displays osu!taiko leaderboard counts, but then I read [this](https://github.com/ppy/osu/blob/master/osu.Game/Scoring/Legacy/ScoreInfoExtensions.cs#L69) and my questions are gone. 
The only confusing part that I found is in [osu!standard simulate command](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/OsuSimulateCommand.cs#L83). I mean this command uses greats (for 300's) and goods (for 100's) which is wrong (if i'm not missing anything), because from [ScoreInfoExtensions](https://github.com/ppy/osu/blob/master/osu.Game/Scoring/Legacy/ScoreInfoExtensions.cs#L92) we know that there are should be Ok's.

smoogipoo:
At one point in time, osu! used to use Great/Good/Meh. These days every non-mania ruleset uses Great/Good/Ok. So the osu! simulate command (and potentially others) is out of date.

minisbett:
I believe this is outdated and implemented by now, so it can be closed.

PR #167, opened 2022-10-17
Add documentation of the current production releases to README

smoogipoo:


peppy:
Would you be against listing all versions across all projects in one place and then linking to that place? Half the reason I wanted this was so we didn't need to make sure it was up-to-date in multiple locations..

smoogipoo:
Nope, perhaps osu-infrastructure is a better place?

peppy:
Sounds good. Then the only question is whether to use a wiki page or commit to the repo. Probably wiki is best?

smoogipoo:
Added to wiki here: https://github.com/ppy/osu-infrastructure/wiki/Star-Rating-and-Performance-Points

I don't think there's much of a difference between commits and wikis, edits to both are trackable (though I guess you can't checkout wikis afaik?).

PR #170, opened 2022-11-29
Improve file selection

stanriders:
Assumes #169 is merged first.

**Before:**  
image
  
**After:**  
image

### DIFF
--- a/PerformanceCalculatorGUI/Components/ExtendedOsuFileSelector.cs
+++ b/PerformanceCalculatorGUI/Components/ExtendedOsuFileSelector.cs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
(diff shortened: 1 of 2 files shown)


PerformanceCalculatorGUI/Components/ExtendedOsuFileSelector.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            // early out to only parse relevant data
+            if (section != Section.Metadata && section != Section.General)
+                return;
+
+            base.ParseLine(beatmap, section, line);
+        }
+    }
+
+    public partial class ExtendedOsuFileSelector : OsuFileSelector
```
peppy, line 25: @smoogipoo 🙈

smoogipoo: Will be fixed with the next SG version (https://github.com/ppy/osu/pull/21451)

peppy: 👍 

let's wait for an `osu.Game` package bump here then. should be a few more days.

smoogipoo (approved):

PR #172, opened 2023-01-10
Move HitObject Panel Information to a Side Bar in the Object Inspector

srb2thepast:
This change simply moves the information contained in each respective mode's panel to a side bar for easier reference.
Fast forwarding or rewinding the timeline with Q and E is added in as well. Holding Shift slows it down more and holding Ctrl speeds it up.

Video Preview:

https://user-images.githubusercontent.com/19241426/211680765-e4eccaf2-6254-4c38-9b0c-937fef31c59a.mp4


stanriders:
I like the concept, but I think the implementation can be simplified. From a quick glance I can see a looot of code quality issues already though, please run code formatting before committing next time

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
```diff
@@ -0,0 +1,182 @@
...
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Containers;
+using osu.Game.Overlays;
+using osuTK;
+
+namespace PerformanceCalculatorGUI.Screens.ObjectInspection
+{
+    public partial class DebugValueList : Container
```
stanriders, line 19: I think it'd be better to call this something like `ObjectDifficultyValuesContainer`, since it's not a list nor its for debugging

PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs
```diff
@@ -32,8 +38,30 @@ public CatchObjectInspectorRuleset(Ruleset ruleset, IBeatmap beatmap, IReadOnlyL
...
+        }
+
+        public void UpdateDebugList(DebugValueList valueList, DifficultyHitObject curDiffHit)
+        {
+            CatchDifficultyHitObject catchDiffHit = (CatchDifficultyHitObject)curDiffHit;
+            string groupName = catchDiffHit.BaseObject.GetType().Name;
+            valueList.AddGroup(groupName, new string[] { "Fruit", "Droplet" });
+            valueList.SetValue(groupName, "Strain Time", catchDiffHit.StrainTime);
+            valueList.SetValue(groupName, "Normalized Position", catchDiffHit.NormalizedPosition);
+            valueList.UpdateValues();
```
stanriders, line 62: This seems to be exactly what the `Bindable`s are for

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
```diff
@@ -0,0 +1,182 @@
...
+                {
+                    t.Scale = new Vector2(1.8f);
+                    t.Font = OsuFont.Torus.With(weight: "Bold");
+                    t.Colour = Colour4.Pink;
+                    t.Shadow = true;
+                });
+
+                foreach (KeyValuePair<string, object> ValuePair in groupDict)
+                {
+                    flowContainer.AddText($"   {ValuePair.Key}:\n", t =>
```
stanriders, line 93: Use `Margin` instead of manually padding using whitespaces

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
```diff
@@ -0,0 +1,182 @@
...
+                new OsuScrollContainer() {
+                    Width = 215,
+                    Height = 670,
+                    ScrollbarAnchor = Anchor.TopLeft,
+                    Child = flowContainer = new TextFlowContainer()
+                    {
+                        Masking = false,
+                        Margin = new MarginPadding { Left = 15 },
+                        Size = new Vector2(200,3500),
+                        Y = 3500,
```
stanriders, line 55: This repeats what's in the `Size` already. Also, you can just use `AutoSizeAxes = Axes.Y`

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
```diff
@@ -0,0 +1,182 @@
...
+                bgBox = new Box
+                {
+                    Colour = colors.Background5,
+                    Alpha = 0.95f,
+                    RelativeSizeAxes = Axes.Y,
+                    Width = 215
+                },
+                new OsuScrollContainer() {
+                    Width = 215,
+                    Height = 670,
```
stanriders, line 48: Why is the height here a const?

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
```diff
@@ -82,28 +88,53 @@ private void load(OverlayColourProvider colourProvider)
...
+            });
+
+            // Object Inspector Container
+            AddInternal(inspectContainer = new Container
+            {
+                Origin = Anchor.Centre,
+                Anchor = Anchor.Centre,
+                Masking = true,
+                CornerRadius = 15f,
+                Padding = new MarginPadding() { Left = side_bar_width, Right = 0 },
```
stanriders, line 115: `Right = 0` can be removed

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
```diff
@@ -82,28 +88,53 @@ private void load(OverlayColourProvider colourProvider)
...
                     {
                         Colour = Colour4.Black,
                         Alpha = 0.95f,
                         RelativeSizeAxes = Axes.Both
                     },
+                }
+            });
+
+            // Object Inspector Container
+            AddInternal(inspectContainer = new Container
```
stanriders, line 109: `AddRange(new[] { all three containers })`

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
```diff
@@ -172,15 +204,34 @@ private void load(OverlayColourProvider colourProvider)
...
+                        new Container{
+                            RelativeSizeAxes = Axes.X,
+                            Height = 3,
+                            Y = 300,
+                            Colour = Colour4.Red,
+                            Children = new Drawable[]{
+                                // shadow
+                                new Circle{Colour = Colour4.Gray, Size = new osuTK.Vector2(10), Y = -3.3f+2, X = -5 },
+                                new Box{ Colour = Colour4.Gray,Size = new osuTK.Vector2(5,20), Y = -8.3f+2, X = 508 },
+                                new Box { Colour = Colour4.Gray, RelativeSizeAxes = Axes.Both, Y = 2},
```
stanriders, line 221: What's going on here?

PerformanceCalculatorGUI/Screens/ObjectInspection/OsuObjectInspectorRuleset.cs
```diff
@@ -19,25 +24,76 @@ namespace PerformanceCalculatorGUI.Screens.ObjectInspection
 {
     public partial class OsuObjectInspectorRuleset : DrawableOsuRuleset
     {
-        public const int HIT_OBJECT_FADE_OUT_EXTENSION = 600;
+        public const int HIT_OBJECT_FADE_OUT_EXTENSION = 0;
```
stanriders, line 27: Just remove it

PerformanceCalculatorGUI/Screens/ObjectInspection/OsuObjectInspectorRuleset.cs
```diff
@@ -19,25 +24,76 @@ namespace PerformanceCalculatorGUI.Screens.ObjectInspection
...
+            var returnedhit = ObjectInspector.GetCurrentHit(Playfield, difficultyHitObjects, lasthit, Clock);
+            if (returnedhit != null)
+            {
+                lasthit = returnedhit;
+                UpdateDebugList(debugValueList, lasthit);
+                // change color
+                var drawHitList = Playfield.AllHitObjects.Where(hit => { return hit.HitObject.StartTime < Clock.CurrentTime; });
+                if (drawHitList.Any())
+                {
+                    drawHitList.Last().Colour = Colour4.Red;
```
stanriders, line 54: I don't think this is necessary

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
```diff
@@ -0,0 +1,182 @@
...
+            foreach (KeyValuePair<string, Dictionary<string, object>> GroupPair in InternalDict)
+            {
+                // Big text
+                string groupName = GroupPair.Key;
+                Dictionary<string, object> groupDict = GroupPair.Value;
+                flowContainer.AddText($"- {GroupPair.Key}\n", t =>
+                {
+                    t.Scale = new Vector2(1.8f);
+                    t.Font = OsuFont.Torus.With(weight: "Bold");
+                    t.Colour = Colour4.Pink;
```
stanriders, line 87: It's not like the tools try to be stylish or anything, but I think this looks kinda ugly

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            flowContainer.Text = "";
+            foreach (KeyValuePair<string, Dictionary<string, object>> GroupPair in InternalDict)
+            {
+                // Big text
+                string groupName = GroupPair.Key;
+                Dictionary<string, object> groupDict = GroupPair.Value;
+                flowContainer.AddText($"- {GroupPair.Key}\n", t =>
+                {
+                    t.Scale = new Vector2(1.8f);
```
stanriders, line 85: Why not just set font size instead?

stanriders (changes requested):
I think the whole logic can be simplified into:

- `ObjectInspector` creates `DebugValueList`
- `DebugValueList` has a `Bindable<DifficultyHitObject>` which is bind to the current object on screen by the `ObjectInspector`

srb2thepast:
> I like the concept, but I think the implementation can be simplified. From a quick glance I can see a looot of code quality issues already though, please run code formatting before committing next time

alright. is running code formatting just done via `dotnet format`?

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
srb2thepast: It was originally due to an issue with sizing, but using more of `AutoSizeAxes` has fixed it. I'm now replacing the two with `AutoSizeAxes = Axes.Both`.

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
srb2thepast: Would it be fine if the color was removed?

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
srb2thepast: I missed this, using font size would be better.

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
srb2thepast: I tried adding a shadow on the bar for the Catch Ruleset for a better look. It resulted in copy pasting the main display then adjusting position and coloring though, which feels a bit messy.

I cant figure out an actual cleaner method to do so. The best thing I can think of is checking if the inspector is ctb then looping & adding, but I feel it's possible for it to be better than that.

PerformanceCalculatorGUI/Screens/ObjectInspection/DebugValueList.cs
srb2thepast: `Margin` seems to apply to each word that gets added rather than only the first. It seems the only other way to do so without whitespace would be to create two Drawables (one `SpriteText` for the text and one `TextFlowContainer` with margin) for every group that's added, but I'm not sure if that's a good direction to go in. Do you have any suggestions?

srb2thepast:
Alright, nearly all the suggestions from each review have been implemented. The ones unresolved are those in which clarification is awaited. I believe it's ready for another review.

PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs
```diff
@@ -21,11 +23,20 @@ public partial class CatchObjectInspectorRuleset : DrawableCatchRuleset
     {
         private readonly CatchDifficultyHitObject[] difficultyHitObjects;
 
-        public CatchObjectInspectorRuleset(Ruleset ruleset, IBeatmap beatmap, IReadOnlyList<Mod> mods, ExtendedCatchDifficultyCalculator difficultyCalculator, double clockRate)
+        [Resolved]
+        private ObjectDifficultyValuesContainer debugValueList { get; set; }
```
stanriders, line 27: Rename

PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs
```diff
@@ -21,11 +23,20 @@ public partial class CatchObjectInspectorRuleset : DrawableCatchRuleset
...
+
+        private DifficultyHitObject lasthit;
+
+        private Bindable<DifficultyHitObject> focusedDiffHitBind;
```
stanriders, line 31: Please don't shorten the variable names
`currentDifficultyHitObjectBindable` reads way better

PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs
```diff
@@ -34,6 +45,34 @@ public CatchObjectInspectorRuleset(Ruleset ruleset, IBeatmap beatmap, IReadOnlyL
...
+        protected override void Update()
+        {
+            base.Update();
+            var hitList = difficultyHitObjects.Where(hit => hit.StartTime < Clock.CurrentTime);
+            if (hitList.Any() && hitList.Last() != lasthit)
+            {
+                lasthit = hitList.Last();
+                focusedDiffHitBind.Value = lasthit;
+            }
+            focusedDiffHitBind.Value = null;
```
stanriders, line 57:
````csharp
focusedDiffHitBind.Value = difficultyHitObjects.LastOrDefault(h => h.StartTime <= Clock.CurrentTime)
````
Storing previous hitobject isn't really necessary as bindable will only update when the value changes

File: PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs [hunk 16]
```diff
@@ -69,14 +108,6 @@ public CatchInspectorDrawableHitObject(CatchDifficultyHitObject dho)
                 [BackgroundDependencyLoader]
                 private void load()
                 {
```
stanriders, line 110: Remove empty functions

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
```diff
@@ -81,62 +88,95 @@ private void load(OverlayColourProvider colourProvider)
...
-            AddInternal(layout = new Container
-            {
-                Origin = Anchor.Centre,
-                Anchor = Anchor.Centre,
-                Masking = true,
-                CornerRadius = 15f,
-                RelativeSizeAxes = Axes.Both,
-                Children = new Drawable[]
+            // Background
+            AddRange(new Drawable[] {
```
stanriders, line 94: Formatting

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
stanriders: Pretty sure this can be done way easier in `CatchObjectInspectorRuleset` without using lots of magic values

PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs
```diff
@@ -34,6 +45,34 @@ public CatchObjectInspectorRuleset(Ruleset ruleset, IBeatmap beatmap, IReadOnlyL
...
+            var hitList = difficultyHitObjects.Where(hit => hit.StartTime < Clock.CurrentTime);
+            if (hitList.Any() && hitList.Last() != lasthit)
+            {
+                lasthit = hitList.Last();
+                focusedDiffHitBind.Value = lasthit;
+            }
+            focusedDiffHitBind.Value = null;
+        }
+
+        public void UpdateDebugList(ObjectDifficultyValuesContainer valueList, DifficultyHitObject curDiffHit)
```
stanriders, line 60: Naming

PerformanceCalculatorGUI/Screens/ObjectInspection/CatchObjectInspectorRuleset.cs
```diff
@@ -34,6 +45,34 @@ public CatchObjectInspectorRuleset(Ruleset ruleset, IBeatmap beatmap, IReadOnlyL
...
+        {
+            if (curDiffHit == null) return;
+
+            CatchDifficultyHitObject catchDiffHit = (CatchDifficultyHitObject)curDiffHit;
+
+            string groupName = catchDiffHit.BaseObject.GetType().Name;
+            valueList.AddGroup(groupName, new string[] { "Fruit", "Droplet" });
+
+            Dictionary<string, Dictionary<string, object>> infoDict = valueList.InfoDictionary.Value;
```
stanriders, line 69: I feel like its still all over the place - I expected hitobject bindable to be passed into the `ObjectDifficultyValuesContainer` and processed there

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectDifficultyValuesContainer.cs
```diff
@@ -0,0 +1,156 @@
...
+    {
+        public Bindable<Dictionary<string, Dictionary<string, object>>> InfoDictionary;
+
+        private Box bgBox;
+        private TextFlowContainer flowContainer;
+        private Bindable<DifficultyHitObject> focusedDiffHit = new();
+
+        public ObjectDifficultyValuesContainer(Bindable<DifficultyHitObject> diffHitBind)
+        {
+            focusedDiffHit.BindTo(diffHitBind);
```
stanriders, line 29: Make `focusedDiffHit` a public get-only property and bind from the outside

stanriders (changes requested):
Still think it can be done way simplier by passing current difficulty hitobject straight into the `ObjectDifficultyValuesContainer`. There are also still a lot of formatting and naming issues

stanriders:
Hope you don't mind, but I've refactored and slightly restyled the PR. Please take a look at what I roughly meant by using a `Bindable` for passing a hitobject into the values container. I'm still not completely satisfied with how it works (ideally I'd get rid of passing everything through rulesets, but that's a task for another day), but it's way more manageable now.

image

stanriders:
> In osu!catch, if the Normalized Position's value is wide enough, it can intersect with the Normalized Position text:

I've changed the font so it fits now 

> While scrolling, the difficulty value text can slightly pass through the top of the ObjectDifficultyValuesContainer.

I'm not sure how to fix that since it looks like a masking issue. I'll leave it to the lazer devs to fix/explain if they want to since its a pretty much nonexisting bug from the usability standpoint

stanriders (approved):

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectDifficultyValuesContainer.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            if (hitObject == null)
+            {
+                hitObjectTypeText.Text = "";
+                return;
+            }
+
+            hitObjectTypeText.Text = hitObject.BaseObject.GetType().Name;
+
+            switch (ruleset.Value.ShortName)
```
peppy, line 106: Could use

```csharp
switch (hitObject)
{
    case OsuDifficultyObject osuObject:
        drawOsuValues(osuObject);
        break;
// etc.
}
```

to avoid string matching.

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
```diff
@@ -227,6 +245,27 @@ protected override bool OnKeyDown(KeyDownEvent e)
                     clock.Start();
             }
 
+            double? seekTo = null;
+
+            if (e.Key == Key.Q)
```
peppy, line 250: Any reason to use non-standard keys? Use arrow keys to match the editor?

srb2thepast: No, arrow keys fit much more.

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
```diff
@@ -247,20 +248,22 @@ protected override bool OnKeyDown(KeyDownEvent e)
...
-                                .HitObjects
-                                .LastOrDefault(x => x.StartTime < clock.CurrentTime)?
-                                .StartTime;
+                seekTo = playableBeatmap.HitObjects
+                                        .LastOrDefault(x => x.StartTime < clock.CurrentTime)?
+                                        .StartTime;
+
+                // slight leeway to make going back beyond just one object possible when the clock is running
+                if (clock.IsRunning)
+                    seekTo -= 100;
```
peppy, line 259: FWIW you're implementing this:

https://github.com/ppy/osu/blob/7bc8908ca9c026fed1d831eb6e58df7624a8d614/osu.Game/Screens/Edit/EditorClock.cs#L87-L92

but in a bit of a weird way. since the keys are skipping to the next object, isn't this potentially going to skip back more than one object at a time? I'm not sure this is a correct approach.

peppy: And please, if changing behaviour like this, please please please don't bundle it in a commit which says nothing about the change being made...

stanriders: I don't think it matters that much to seek accurately when the clock is running anyway, the only real use case for it I can come up with is going back like a couple of seconds using arrows instead of the timeline. Since you can only see one object's difficulty values at a time, having clock running makes the inspection pretty much unusable anyway

Sure, I'll add minor stuff to the commit messages too

peppy (approved):

PerformanceCalculatorGUI/Screens/ObjectInspection/ObjectInspector.cs
peppy: Please fix that in a follow-up PR. This one already has too much going on.

PR #195, opened 2024-01-30
Add support for computing performance of non-legacy scores

smoogipoo:
Shouldn't conflict (apart from `using`s) with https://github.com/ppy/osu-tools/pull/194 .

`LegacyHelper` implementation taken from https://github.com/ppy/osu-queue-score-statistics/blob/master/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/LegacyModsHelper.cs Not sure if this needs to be moved into `osu.Game` at this point.

Commands:
```
performance score <score-id> <client-id> <client-secret> [-a|--online-attributes]
performance legacy-score <score-id> <ruleset-id> <client-id> <client-secret> [-a|--online-attributes]
```

Note that the order of arguments in the `legacy-score` command differs from `master` in which it is currently `<ruleset-id> <score-id>`.

This also adds support for the `-a|--online-attributes` option, which allows database attributes to be used instead by querying the `beatmap/{id}/attributes` endpoint.

Can be tested with:
```diff
diff --git a/PerformanceCalculator/Program.cs b/PerformanceCalculator/Program.cs
index 42f4b4b..3d662db 100644
--- a/PerformanceCalculator/Program.cs
+++ b/PerformanceCalculator/Program.cs
@@ -26,7 +26,7 @@ namespace PerformanceCalculator
     [HelpOption("-?|-h|--help")]
     public class Program
     {
-        public static readonly EndpointConfiguration ENDPOINT_CONFIGURATION = new ProductionEndpointConfiguration();
+        public static readonly EndpointConfiguration ENDPOINT_CONFIGURATION = new ExperimentalEndpointConfiguration();

         public static void Main(string[] args)
         {
```

### DIFF
--- a/PerformanceCalculator/LegacyHelper.cs
+++ b/PerformanceCalculator/LegacyHelper.cs
@@ -2,20 +2,20 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using osu.Framework.Audio.Track;
-using osu.Framework.Graphics.Textures;
 using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.Legacy;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Catch;
+using osu.Game.Rulesets.Catch.Difficulty;
+using osu.Game.Rulesets.Difficulty;
 using osu.Game.Rulesets.Mania;
+using osu.Game.Rulesets.Mania.Difficulty;
 using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.Osu;
+using osu.Game.Rulesets.Osu.Difficulty;
 using osu.Game.Rulesets.Taiko;
-using osu.Game.Skinning;
-using osu.Game.Utils;
+using osu.Game.Rulesets.Taiko.Difficulty;
 
 namespace PerformanceCalculator
 {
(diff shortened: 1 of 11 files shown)


PerformanceCalculator/LegacyHelper.cs
```diff
@@ -63,51 +63,73 @@ public static string GetRulesetShortNameFromId(int id)
...
-        /// <summary>
-        /// Transforms a given <see cref="Mod"/> combination into one which is applicable to legacy scores.
-        /// This is used to match osu!stable/osu!web calculations for the time being, until such a point that these mods do get considered.
-        /// </summary>
-        public static Mod[] ConvertToLegacyDifficultyAdjustmentMods(BeatmapInfo beatmapInfo, Ruleset ruleset, Mod[] mods)
+        public const LegacyMods KEY_MODS = LegacyMods.Key1 | LegacyMods.Key2 | LegacyMods.Key3 | LegacyMods.Key4 | LegacyMods.Key5 | LegacyMods.Key6 | LegacyMods.Key7 | LegacyMods.Key8
+                                           | LegacyMods.Key9 | LegacyMods.KeyCoop;
+
+        // See: https://github.com/ppy/osu-performance/blob/83c02f50315a4ef7feea80acb84c66ee437d7210/include/pp/Common.h#L109-L129
+        public static LegacyMods MaskRelevantMods(LegacyMods mods, bool isConvertedBeatmap, int rulesetId)
```
bdach, line 70: Could at least link back to the [`osu-queue-score-statistics` version of this](https://github.com/ppy/osu-queue-score-statistics/blob/2264bfa68e14bb16ec71a7cac2072bdcfaf565b6/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/LegacyModsHelper.cs#L16).

bdach (changes requested):
as above

bdach (approved):
think this can be merged now, https://github.com/ppy/osu/pull/27069 can bubble down here whenever most convenient

PR #196, opened 2024-02-02
Update to .NET 8 + R# 2023.3.3

smoogipoo:


peppy:
At some point we'll need to silence the hints/whatever from inspections for primary ctors etc.

peppy (approved):

PR #198, opened 2024-02-14
Reworked hit result generator based on accuracy

Givikap120:
It used to just spam 50s until it reach target accuracy
Now it's actually trying to fit it in realistic way


tsunyoku (changes requested):
Would be nice to add explanation on what this is doing and why. All I see is logic with no clear reasoning on why it does what it does.

Givikap120:
> Would be nice to add explanation on what this is doing and why. All I see is logic with no clear reasoning on why it does what it does.

Added very detailed explanation of what I'm doing here. You can remove unnecessary things.

stanriders (changes requested):
Please copy your implementation to the CLI perfcalc as well: https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/OsuSimulateCommand.cs#L63

Givikap120:
> Please copy your implementation to the CLI perfcalc as well: https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/OsuSimulateCommand.cs#L63

Fixed

stanriders (changes requested):
Please fix CI issues

stanriders:
image

smoogipoo:
CodeFactor is fine, don't mind that too much.

stanriders (approved):
I'd say a lot of this is redundant (does it really matter if scores below 25% accuracy are accurately calculated?..) but I guess it doesnt hurt to have

Issue #201, opened 2024-03-14
Add buttons that can open the file path directly.


CloudeaSoft:
As it shown in the below picture, but I can't make it look good. Can anyone help me? ^_^
![{7D93E44B-7752-457f-8236-A1EF398EC221}](https://github.com/ppy/osu-tools/assets/129159758/5319ce71-5cd6-472c-a266-017b5eeef14a)

CloudeaSoft:
The codes are in [https://github.com/ppy/osu-tools/pull/202](url)

PR #202, opened 2024-03-14
Add buttons that can open the file path directly.

CloudeaSoft:
[https://github.com/ppy/osu-tools/issues/201](url)
The codes for the issue#201.

### DIFF
--- a/PerformanceCalculatorGUI/Components/SettingsPopover.cs
+++ b/PerformanceCalculatorGUI/Components/SettingsPopover.cs
@@ -10,6 +10,7 @@ using osu.Framework.Extensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
+using osu.Framework.Platform;
 using osu.Game.Configuration;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Containers;
@@ -27,6 +28,8 @@ namespace PerformanceCalculatorGUI.Components
 
         private LinkFlowContainer linkContainer;
 
+        private GameHost gameHost;
+
         private Bindable<string> clientIdBindable;
         private Bindable<string> clientSecretBindable;
         private Bindable<string> pathBindable;
@@ -36,7 +39,7 @@ namespace PerformanceCalculatorGUI.Components
         private const string api_key_link = "https://osu.ppy.sh/home/account/edit#new-oauth-application";
 
         [BackgroundDependencyLoader]
-        private void load(SettingsManager configManager, OsuConfigManager osuConfig)
+        private void load(SettingsManager configManager, OsuConfigManager osuConfig, GameHost gameHost)
         {
             this.configManager = configManager;
             clientIdBindable = configManager.GetBindable<string>(Settings.ClientId);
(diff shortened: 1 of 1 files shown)


PerformanceCalculatorGUI/Components/SettingsPopover.cs
```diff
@@ -134,5 +171,20 @@ private void saveConfig()
...
+        }
+
+        private void openCacheDirectory()
+        {
+            openFolder(cacheBindable.Value);
+        }
+
+        private static void openFolder(string folderPath)
+        {
+            Process.Start("explorer.exe", folderPath);
```
stanriders, line 187: This wont work on platforms other than windows

PerformanceCalculatorGUI/Components/SettingsPopover.cs
```diff
@@ -83,17 +84,53 @@ private void load(SettingsManager configManager, OsuConfigManager osuConfig)
...
+                                            Current = { BindTarget = pathBindable }
+                                        },
+                                        new RoundedButton
+                                        {
+                                            Anchor = Anchor.CentreLeft,
+                                            Origin = Anchor.CentreLeft,
+                                            Size = new Vector2(100, 40),
+                                            Margin = new MarginPadding { Left = 10 },
+                                            Text = "Open",
+                                            Action = openPathDirectory
```
stanriders, line 110: I'm not sure how useful this even is - you can always just copypaste the path from the text field into the file manager address bar

stanriders (changes requested):

PerformanceCalculatorGUI/Components/SettingsPopover.cs
CloudeaSoft: Ok, i'll try to make it more simply for other platforms.

PerformanceCalculatorGUI/Components/SettingsPopover.cs
CloudeaSoft: I thought osu! have buttons for root path/skin path/beatmap path(in editor), so i tried to add this button.

PerformanceCalculatorGUI/Components/SettingsPopover.cs
```diff
@@ -134,5 +173,39 @@ private void saveConfig()
...
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                // For Linux
+                Process.Start("xdg-open", path);
+            }
+            else
+            {
+                // For others, do nothing
+            }
+        }
```
smoogipoo: The correct solution here is to use `Storage.PresentExternally()`. You should be able to directly inject `GameHost` as a dependency here. Something like the following might be enough:
```
host.GetStorage(path).PresentExternally();
```

smoogipoo (changes requested):

PerformanceCalculatorGUI/Components/SettingsPopover.cs
CloudeaSoft: Thx for the advice! :)

PerformanceCalculatorGUI/Components/SettingsPopover.cs
Ireadosucode: it's quality of life for sure- not to mention a lot of people don't know how to do that

Ireadosucode: somehow..

PR #203, opened 2024-04-24
Fix mania GenerateHitResults() bugs

Syriiin:
There are a couple bugs:

1. An odd number of `Perfect`/`Great` hits would miss a hit and throw off following calculations.
2. `Ok`s were duplicated as `Meh`s

This function is quite hard to parse so it's not surprising these slipped in.
I think these variables could do with some renaming, but that's beyond the scope of this fix.

On a side note, my gut says it should just assume all `Great`s are `Perfect`s rather than splitting them down the middle.
Usually one would expect that omitting a parameter would set it as the best possible value.

### DIFF
--- a/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
@@ -80,11 +80,12 @@ namespace PerformanceCalculator.Simulate
             int delta = targetTotal - remainingHits;
 
             // Each great and perfect increases total by 5 (great-meh=5)
-            // There is no difference in accuracy between them, so just halve arbitrarily.
-            greats = Math.Min(delta / 5, remainingHits) / 2;
-            int perfects = greats.Value;
+            // There is no difference in accuracy between them, so just halve arbitrarily (favouring perfects for an odd number).
+            int greatsAndPerfects = Math.Min(delta / 5, remainingHits);
+            greats = greatsAndPerfects / 2;
+            int perfects = greats.Value + greatsAndPerfects % 2;
             delta -= (greats.Value + perfects) * 5;
-            remainingHits -= (greats.Value + perfects);
+            remainingHits -= greats.Value + perfects;
 
             // Each good increases total by 3 (good-meh=3).
             countGood = Math.Min(delta / 3, remainingHits);
@@ -93,6 +94,7 @@ namespace PerformanceCalculator.Simulate
 
             // Each ok increases total by 1 (ok-meh=1).
             oks = delta;
+            remainingHits -= oks.Value;
 
             // Everything else is a meh, as initially assumed.
             countMeh = remainingHits;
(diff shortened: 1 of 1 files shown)


PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
```diff
@@ -80,11 +80,12 @@ public class ManiaSimulateCommand : SimulateCommand
             int delta = targetTotal - remainingHits;
 
             // Each great and perfect increases total by 5 (great-meh=5)
-            // There is no difference in accuracy between them, so just halve arbitrarily.
-            greats = Math.Min(delta / 5, remainingHits) / 2;
-            int perfects = greats.Value;
+            // There is no difference in accuracy between them, so just halve arbitrarily (favouring perfects for an odd number).
+            int greatsAndPerfects = Math.Min(delta / 5, remainingHits);
+            greats = greatsAndPerfects / 2;
+            int perfects = greats.Value + (greatsAndPerfects % 2);
```
bdach, line 86: I would have done

```suggestion
            int perfects = greatsAndPerfects - greats.Value;
```

to dispel any and all doubts of inaccuracy. I guess what you have works but feels needlessly contrived?

bdach:
> On a side note, my gut says it should just assume all `Great`s are `Perfect`s rather than splitting them down the middle. Usually one would expect that omitting a parameter would set it as the best possible value.

The reason I did this originally is that hitting all perfects seemed - excluding the easiest maps - rather unrealistic. Usually when someone isn't providing exact difficulty numbers they're providing raw accuracy so extrapolations need to be made.

And yeah that wouldn't matter one bit in score V1 but it does actually matter a bit now that perfects are scored 305 rather than 300 when it comes to accuracy at least.

PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
Syriiin: Personally I think divmod is a little more clear since the fact that it's integer division isn't immediately obvious.
But simply subtracting from that value is certainly more readable given c# doesn't have a built in divmod function.
I'll change it tonight :ok_hand:

Syriiin:
> The reason I did this originally is that hitting all perfects seemed - excluding the easiest maps - rather unrealistic. Usually when someone isn't providing exact difficulty numbers they're providing raw accuracy so extrapolations need to be made.

Yea in practice for real plays it makes sense, it's just the expectation I think most people would have from difficulty calculators though.
To get the max pp from a score, you just give the beatmap with no parameters. :man_shrugging:
Needing to explicitly pass a `--greats 0` (especially when it's divergent from existing patterns) just feels wrong.
If someone explicitly passed accuracy as `--accuracy 99` or something, I can see splitting the difference between greats and perfects as more justifiable there.
Anyway, off topic!

bdach (approved):

PR #204, opened 2024-04-29
Fix online attribute flag in score performance command not retrieving full difficulty attributes

bdach:
Found when investigating https://github.com/ppy/osu/issues/28006.

If the `performance score` command is given the `-a` flag, it is supposed to use the current online beatmap difficulty attributes. It does so by querying API, [which queries `osu-beatmap-difficulty-cache`.](https://github.com/ppy/osu-web/blob/0444f7772623f3ee4941065da7bdb0e29adfbe01/app/Libraries/BeatmapDifficultyAttributes.php#L14)

Problem is, that some of the difficulty attributes, namely the ones that are already present on `APIBeatmap`, [are not serialised out by `osu-beatmap-difficulty-cache`](https://github.com/ppy/osu-beatmap-difficulty-lookup-cache/blob/db2203368221109803f2031788da31deb94e0f11/BeatmapDifficultyLookupCache/DifficultyCache.cs#L125-L128).

`osu-tools` is blissfully unaware of this, and as such attempts to deserialise the difficulty attributes returned from cache-then-web raw, which leads to it just using the default values (e.g. zeroes) for attributes that the difficulty lookup cache does not serialise out. This leads to obviously bogus results.

To fix, retrieve the `APIBeatmap` manually and do some (admittedly gnarly) grafting in order to produce a fully populated object. It's a bit ugly but it works locally without further changes so maybe fine?

The alternative would be to fix this on the beatmap difficulty cache, I guess. That said it's not like the cache/API returns these fields wrong either, it does correctly omit them, which is why I'm trying this first.

### DIFF
--- a/PerformanceCalculator/Performance/ScorePerformanceCommand.cs
+++ b/PerformanceCalculator/Performance/ScorePerformanceCommand.cs
@@ -84,22 +84,40 @@ namespace PerformanceCalculator.Performance
                 { "mods", ((int)mods).ToString(CultureInfo.InvariantCulture) }
             };
 
+            var beatmap = GetJsonFromApi<APIBeatmap>($"beatmaps/{beatmapId}");
+
             switch (rulesetId)
             {
                 case 0:
-                    return GetJsonFromApi<AttributesResponse<OsuDifficultyAttributes>>($"beatmaps/{beatmapId}/attributes", HttpMethod.Post, parameters).Attributes;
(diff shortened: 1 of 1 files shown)


peppy (approved):
Seems okay if we want to avoid serialising these in `osu-beatmap-difficulty-cache`. Probably best until there's another usage of them.

smoogipoo (approved):

PR #208, opened 2024-07-18
Change ObjectInspector to work via object selection

Givikap120:
Known issues:

- Taiko selected objects are appearing as "deselected" when scrolling back earlier than half of the playfield
- Taiko selected objects are appearing as "deselected" after time it should be hit (even tho Drawable looks like it's still goes to the left)
- Catch selected objects that were created from nested fruits (slider ends) are fading-in for some reason

First two are most likely because of how pooling in Taiko works
Third is because of some internal catch playfield behavior (didn't found where the fade is applied)

Don't know how to fix the issues, and they're not critical, so waiting on commentary of this situation


stanriders:
> * Do I need to merge STD classes with abstract base as the abstractions isn't used anywhere else?

Yes
> * How can I add selection logic to the `CatchSelectableDrawableObject` and `TaikoSelectableDrawableObject` without copying? Or I just need to copy it and that's all?

Copying is fine if can't be moved out easily
> * Do I need to implement selectable spinner (and other objects without respective difficulty hit objects)?

No
> * Selectable object creation can be done with using DrawablePool. But seems to work fine without it. Do I need to do it?

Try running very high object count maps and see how big of a performance hit you get. If it's negligible and your system is more or less average it should be fine I think

Givikap120:
Known issues:
- Taiko selected objects are appearing as "deselected" when scrolling back earlier than half of the playfield
- Taiko selected objects are appearing as "deselected" after time it should be hit (even tho Drawable looks like it's still goes to the left)
- Catch selected objects that were creating from nested fruits (slider ends) are fading-in for some reason

First two are most likely because of how pooling in Taiko works
Third is because of some internal catch playfield behavior (didn't found where the fade is applied)

Don't know how to fix the issues, and they're not critical, so waiting on commentary of this situation

PerformanceCalculatorGUI/Screens/ObjectInspection/Catch/CatchObjectInspectorRuleset.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            protected override bool OnClick(ClickEvent e)
+            {
+                if (e.Button == MouseButton.Right)
+                    return false;
+
+                CatchSelectableHitObject newSelectedObject = null;
+
+                // This search can be long if list of objects is very big. Potential for optimization
+                foreach (var selectable in selectables)
```
stanriders, line 140: Shouldn't the selectable itself process the input and playfield (or inspection panel) listen to select events? I feel like iterating over every object might get expensive very fast

stanriders (changes requested):

PerformanceCalculatorGUI/Screens/ObjectInspection/Catch/CatchObjectInspectorRuleset.cs
Givikap120: Fixed

PerformanceCalculatorGUI/Screens/ObjectInspection/Catch/CatchObjectInspectorRuleset.cs
```diff
@@ -0,0 +1,115 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+#nullable enable
```
smoogipoo, line 4: In general we don't `#nullable enable` in new files.

stanriders: I'm planning to go over the whole project and enabling nullables everywhere after I'm done refactoring this pr

PR #209, opened 2024-07-20
Added hidden to relevant mods

MrHeliX:
Fixes hidden not counted as relevant mods when using simulate command without the `-nc` parameter

Before:
WindowsTerminal_2024-07-20_17-18-01

After:
WindowsTerminal_2024-07-20_17-18-14

### DIFF
--- a/PerformanceCalculator/LegacyHelper.cs
+++ b/PerformanceCalculator/LegacyHelper.cs
@@ -67,6 +67,7 @@ namespace PerformanceCalculator
                                            | LegacyMods.Key9 | LegacyMods.KeyCoop;
 
         // See: https://github.com/ppy/osu-queue-score-statistics/blob/2264bfa68e14bb16ec71a7cac2072bdcfaf565b6/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/LegacyModsHelper.cs
+        // Since hidden always affects pp calculation, it has been added to the base relevant mods, contrary to the source of this method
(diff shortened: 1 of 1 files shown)


minisbett:
You would want to remove hidden [here](https://github.com/ppy/osu-tools/blob/3a1d088bdf2e3f31b5970fb8f74c5ab93ac36d2e/PerformanceCalculator/LegacyHelper.cs#L78) then, the code originally intended to only consider hidden a relevant mod if flashlight is enabled

MrHeliX:
> You would want to remove hidden [here](https://github.com/ppy/osu-tools/blob/3a1d088bdf2e3f31b5970fb8f74c5ab93ac36d2e/PerformanceCalculator/LegacyHelper.cs#L78) then, the code originally intended to only consider hidden a relevant mod if flashlight is enabled
> 
> It'd also potentially make sense to make an in-line comment for it, as this method would now differ from the source for this method specified in the comment above

Of course yeah, updated PR

smoogipoo:
I'm not sure this is entirely correct because SR-wise HD only has an effect when combined with FL.

That is, perhaps the following should only be applying this filter to the mods passed to the difficulty calculator, rather than the `ScoreInfo`:
https://github.com/ppy/osu-tools/blob/a1bb55f1008e9804b740769c139dbec67ac7f4a9/PerformanceCalculator/Simulate/SimulateCommand.cs#L59-L77

As it is, it also means some other mods like TC are filtered out.

MrHeliX:
Closing this in favour of https://github.com/ppy/osu-tools/pull/216

PR #210, opened 2024-07-20
Fix Strain Graph lag on long maps

Givikap120:
The problem is that every strain graph bar have tooltip. This means that if map have 8k strains - we have 8000 * 4 skills = 24000 tooltips, causing intensive lag
In this PR I'm separating strain bars from tooltips, and making a fixed amount of tooltip bars (I picked 1000 as it's precise enough while not causing lag)
Also, I packed some actions into the function as the function was so big that It was difficult to orient in it


smoogipoo (approved):
Seems okay

PR #213, opened 2024-08-13
Implemented functionality to calculate all local scores using Realm

Givikap120:
Now Profile page will have a switch that can change mode of calculation to local.
Local mode has additional setting menu.
As local mode doesn't have deltas - `ExtendedProfileScore` was divided into two: `ProfileScore` and `ExtendedProfileScore`, where only second has delta display.
Calculation of 50k scores will be slow, that's why new caching algorithm was implemented. This caching will be also used in `BeatmapLeaderboardScreen`, what will speed-up calculation of leaderboard on long maps in around 10 times.

### DIFF
--- a/PerformanceCalculatorGUI/Components/ExtendedProfileScore.cs
+++ b/PerformanceCalculatorGUI/Components/ExtendedProfileScore.cs
@@ -1,423 +1,115 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
-using System.Collections.Generic;
 using System.Linq;
-using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
-using osu.Framework.Graphics.Shapes;
-using osu.Framework.Input.Events;
-using osu.Framework.Localisation;
-using osu.Framework.Platform;
-using osu.Game.Beatmaps;
 using osu.Game.Graphics;
-using osu.Game.Graphics.Containers;
 using osu.Game.Graphics.Sprites;
 using osu.Game.Online.API.Requests.Responses;
-using osu.Game.Online.Leaderboards;
-using osu.Game.Overlays;
-using osu.Game.Overlays.Profile.Sections;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Difficulty;
-using osu.Game.Rulesets.Scoring;
-using osu.Game.Rulesets.UI;
-using osu.Game.Utils;
-using osuTK;
-using PerformanceCalculatorGUI.Components.TextBoxes;
 
 namespace PerformanceCalculatorGUI.Components
 {
-    public class ExtendedScore
+    public class ExtendedProfileScore : ProfileScore
     {
-        public SoloScoreInfo SoloScore { get; }
         public double LivePP { get; }
-
-        public Bindable<int> Position { get; } = new Bindable<int>();
         public Bindable<int> PositionChange { get; } = new Bindable<int>();
 
-        public PerformanceAttributes PerformanceAttributes { get; }
-
-        public ExtendedScore(SoloScoreInfo score, double livePP, PerformanceAttributes attributes)
+        public ExtendedProfileScore(SoloScoreInfo score, double livePP, PerformanceAttributes attributes)
+            : base(score, attributes)
         {
-            SoloScore = score;
-            PerformanceAttributes = attributes;
             LivePP = livePP;
         }
     }
 
-    public partial class ExtendedProfileItemContainer : ProfileItemContainer
+    public partial class DrawableExtendedProfileScore : DrawableProfileScore
     {
-        public Action OnHoverAction { get; set; }
-        public Action OnUnhoverAction { get; set; }
+        protected new ExtendedProfileScore Score { get; }
 
-        public ExtendedProfileItemContainer()
-        {
-            CornerRadius = ExtendedLabelledTextBox.CORNER_RADIUS;
-        }
-
-        protected override bool OnHover(HoverEvent e)
-        {
-            OnHoverAction?.Invoke();
-            return base.OnHover(e);
-        }
-
-        protected override void OnHoverLost(HoverLostEvent e)
-        {
-            OnUnhoverAction?.Invoke();
-            base.OnHoverLost(e);
-        }
-    }
-
-    public partial class ExtendedProfileScore : CompositeDrawable
-    {
-        private const int height = 40;
-        private const int performance_width = 100;
-        private const int rank_difference_width = 35;
-        private const int small_text_font_size = 11;
-
-        private const float performance_background_shear = 0.45f;
-
-        protected readonly ExtendedScore Score;
-
-        [Resolved]
-        private OsuColour colours { get; set; }
-
-        [Resolved]
-        private OverlayColourProvider colourProvider { get; set; }
-
-        private OsuSpriteText positionChangeText;
-
-        public ExtendedProfileScore(ExtendedScore score)
+        public DrawableExtendedProfileScore(ExtendedProfileScore score)
+            : base(score)
         {
             Score = score;
-
-            RelativeSizeAxes = Axes.X;
-            Height = height;
         }
 
-        [BackgroundDependencyLoader]
-        private void load(RulesetStore rulesets)
+        protected override void LoadComplete()
         {
-            AddInternal(new ExtendedProfileItemContainer
-            {
-                OnHoverAction = () =>
-                {
-                    positionChangeText.Text = $"#{Score.Position.Value}";
-                },
-                OnUnhoverAction = () =>
-                {
-                    positionChangeText.Text = $"{Score.PositionChange.Value:+0;-0;-}";
-                },
-                Children = new Drawable[]
-                {
-                    new Container
-                    {
-                        Name = "Rank difference",
-                        RelativeSizeAxes = Axes.Y,
-                        Anchor = Anchor.CentreLeft,
-                        Origin = Anchor.CentreLeft,
-                        Width = rank_difference_width,
-                        Child = positionChangeText = new OsuSpriteText
-                        {
-                            Anchor = Anchor.Centre,
-                            Origin = Anchor.Centre,
-                            Colour = colourProvider.Light1,
-                            Text = Score.PositionChange.Value.ToString()
-                        }
-                    },
-                    new Container
-                    {
-                        Name = "Score info",
-                        RelativeSizeAxes = Axes.Both,
-                        Padding = new MarginPadding { Left = rank_difference_width, Right = performance_width },
-                        Children = new Drawable[]
-                        {
-                            new FillFlowContainer
-                            {
-                                Anchor = Anchor.CentreLeft,
-                                Origin = Anchor.CentreLeft,
-                                AutoSizeAxes = Axes.Both,
-                                Direction = FillDirection.Horizontal,
-                                Spacing = new Vector2(10, 0),
-                                Children = new Drawable[]
-                                {
-                                    new UpdateableRank(Score.SoloScore.Rank)
-                                    {
-                                        Anchor = Anchor.CentreLeft,
-                                        Origin = Anchor.CentreLeft,
-                                        Size = new Vector2(50, 20),
-                                    },
-                                    new FillFlowContainer
-                                    {
-                                        Anchor = Anchor.CentreLeft,
-                                        Origin = Anchor.CentreLeft,
-                                        AutoSizeAxes = Axes.Both,
-                                        Direction = FillDirection.Vertical,
-                                        Spacing = new Vector2(0, 0.5f),
-                                        Children = new Drawable[]
-                                        {
-                                            new ScoreBeatmapMetadataContainer(Score.SoloScore.Beatmap),
-                                            new FillFlowContainer
-                                            {
-                                                AutoSizeAxes = Axes.Both,
-                                                Direction = FillDirection.Horizontal,
-                                                Spacing = new Vector2(15, 0),
-                                                Children = new Drawable[]
-                                                {
-                                                    new OsuSpriteText
-                                                    {
-                                                        Text = $"{Score.SoloScore.Beatmap?.DifficultyName}",
-                                                        Font = OsuFont.GetFont(size: 12, weight: FontWeight.Regular),
-                                                        Colour = colours.Yellow
-                                                    },
-                                                    new DrawableDate(Score.SoloScore.EndedAt, 12)
-                                                    {
-                                                        Colour = colourProvider.Foreground1
-                                                    }
-                                                }
-                                            }
-                                        }
-                                    }
-                                }
-                            },
-                            new FillFlowContainer
-                            {
-                                Anchor = Anchor.CentreRight,
-                                Origin = Anchor.CentreRight,
-                                AutoSizeAxes = Axes.X,
-                                RelativeSizeAxes = Axes.Y,
-                                Direction = FillDirection.Horizontal,
-                                Children = new Drawable[]
-                                {
-                                    new Container
-                                    {
-                                        AutoSizeAxes = Axes.X,
-                                        RelativeSizeAxes = Axes.Y,
-                                        Padding = new MarginPadding { Horizontal = 10, Vertical = 5 },
-                                        Anchor = Anchor.CentreRight,
-                                        Origin = Anchor.CentreRight,
-                                        Child = new FillFlowContainer
-                                        {
-                                            AutoSizeAxes = Axes.Both,
-                                            Direction = FillDirection.Vertical,
-                                            Origin = Anchor.CentreLeft,
-                                            Anchor = Anchor.CentreLeft,
-                                            Children = new Drawable[]
-                                            {
-                                                new FillFlowContainer
-                                                {
-                                                    AutoSizeAxes = Axes.Both,
-                                                    Direction = FillDirection.Horizontal,
-                                                    Spacing = new Vector2(10, 0),
-                                                    Children = new Drawable[]
-                                                    {
-                                                        new FillFlowContainer
-                                                        {
-                                                            Anchor = Anchor.Centre,
-                                                            Origin = Anchor.Centre,
-                                                            Width = 110,
-                                                            RelativeSizeAxes = Axes.Y,
-                                                            Direction = FillDirection.Vertical,
-                                                            Children = new Drawable[]
-                                                            {
-                                                                new OsuSpriteText
-                                                                {
-                                                                    Text = Score.SoloScore.Accuracy.FormatAccuracy(),
-                                                                    Font = OsuFont.GetFont(weight: FontWeight.Bold, italics: true),
-                                                                    Colour = colours.Yellow,
-                                                                    Anchor = Anchor.TopCentre,
-                                                                    Origin = Anchor.TopCentre
-                                                                },
-                                                                new OsuSpriteText
-                                                                {
-                                                                    Text = $"{Score.SoloScore.MaxCombo}x {{ {formatStatistics(Score.SoloScore.Statistics)} }}",
-                                                                    Font = OsuFont.GetFont(size: small_text_font_size, weight: FontWeight.Regular),
-                                                                    Colour = colourProvider.Light2,
-                                                                    Anchor = Anchor.TopCentre,
-                                                                    Origin = Anchor.TopCentre
-                                                                },
-                                                            }
-                                                        },
-                                                        new FillFlowContainer
-                                                        {
-                                                            Anchor = Anchor.Centre,
-                                                            Origin = Anchor.Centre,
-                                                            Width = 60,
-                                                            AutoSizeAxes = Axes.Y,
-                                                            Direction = FillDirection.Vertical,
-                                                            Children = new Drawable[]
-                                                            {
-                                                                new Container
-                                                                {
-                                                                    AutoSizeAxes = Axes.Y,
-                                                                    Child = new OsuSpriteText
-                                                                    {
-                                                                        Font = OsuFont.GetFont(weight: FontWeight.Bold),
-                                                                        Text = $"{Score.LivePP:0}pp"
-                                                                    },
-                                                                },
-                                                                new OsuSpriteText
-                                                                {
-                                                                    Font = OsuFont.GetFont(size: small_text_font_size),
-                                                                    Text = "live"
-                                                                }
-                                                            }
-                                                        }
-                                                    }
-                                                }
-                                            }
-                                        }
-                                    },
-                                    new FillFlowContainer
-                                    {
-                                        AutoSizeAxes = Axes.Both,
-                                        Anchor = Anchor.CentreRight,
-                                        Origin = Anchor.CentreRight,
-                                        Direction = FillDirection.Horizontal,
-                                        Spacing = new Vector2(2),
-                                        Children = Score.SoloScore.Mods.Select(mod =>
-                                        {
-                                            var ruleset = rulesets.GetRuleset(Score.SoloScore.RulesetID) ?? throw new InvalidOperationException();
+            base.LoadComplete();
 
-                                            return new ModIcon(ruleset.CreateInstance().CreateModFromAcronym(mod.Acronym)!)
-                                            {
-                                                Scale = new Vector2(0.35f)
-                                            };
-                                        }).ToList(),
-                                    }
-                                }
-                            }
-                        }
-                    },
-                    new Container
+            Score.Position.UnbindEvents();
+            Score.PositionChange.BindValueChanged(v => { PositionText.Text = $"{v.NewValue:+0;-0;-}"; });
+        }
+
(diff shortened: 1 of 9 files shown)


Natelytle:
image
Parenthesis are useless imo (+ the wording could be better, "for example" and "anyway" isn't wording you should use in a context like this.)
Scorev1'd scores and unsubmitted scores should be combined into one option. If you do decide to keep it, fix the wording, something like "Only calculate highest legacy score".
I don't get why RX is excluded? Autopilot I understand, but RX is a very weird choice since it is handled in PP now.

As well, the calculation seems to fail for me. Not sure why, no error message.

Givikap120:
> image Parenthesis are useless imo (+ the wording could be better, "for example" and "anyway" isn't wording you should use in a context like this.) Scorev1'd scores and unsubmitted scores should be combined into one option. If you do decide to keep it, fix the wording, something like "Only calculate highest legacy score". I don't get why RX is excluded? Autopilot I understand, but RX is a very weird choice since it is handled in PP now.
> 
> As well, the calculation seems to fail for me. Not sure why, no error message.

Scorev1-ed scores are submitted (as well as scores on graveyard maps are submitted too). So it doesn't makes sense to filter-out those scores if we ignore not submitted scores.
RX is excluded because RX was overweight with CSR, making my #1 unranked score to be 300pp on 8 star with RX. But that's silly reason, ye, I should probably remove it.
Also >14 star unranked maps was excluded in the code because aspire map sliderend miss penalty is super bad. Don't sure should I keep this or no.
And your screenshot seems to have old text, scorev1-ning option name is "Enable Scorev1 score overwrite for legacy scores".
About linux - idk how to deal with it. I will let StanR to look into this issue.

smoogipoo:
What is the use-case of this - why compute from local scores rather than online profile?

stanriders:
> What is the use-case of this - why compute from local scores rather than online profile?

The idea is to be able to compute more scores per profile since API is limited to 100 only and database dumps are limited to top 10k players

Givikap120:
> What is the use-case of this - why compute from local scores rather than online profile?

- checking your real profile with CSR is pretty impossible now - this thing fixes it
- also it's useful for players who often play graveryard maps and have a lot of top plays on them (example - Plasma, Lolu)
- + players like me, who have a lot of RA scores - I can finally see where they should be
- it's just funny in general to see different statistics like "amount of 200pp plays you have" or "how much pp you would have only on unranked maps"
- in future I will also add mod filter, so you will be able to calculate your top plays only for NM for example
- precedent for having Realm access, meaning that you now can have simulate score on lazer beatmap without having jump through hoops with exporting it back to stable. Also it means proper search in the simulate menu. Also it means you can use beatmaps directly from lazer, without having to download it into local storage, costing time and space.

I hope this is enough reasons to understand why this PR is useful

minisbett:
Selecting maps from lazer instead of having to bother with .osu files? yes please merge this that'd be sooo useful

stanriders:
> Selecting maps from lazer instead of having to bother with .osu files? yes please merge this that'd be sooo useful

This is not what it does. It's only allowing you to calculate already existing scores from your lazer scores database. You can also just use beatmap ids instead of .osu files already

PerformanceCalculatorGUI/Screens/BeatmapLeaderboardScreen.cs
```diff
@@ -227,14 +232,18 @@ private void calculate()
...
                     var scoreInfo = score.ToScoreInfo(rulesets, working.BeatmapInfo);
+                    Mod[] mods = score.Mods.Select(x => x.ToMod(rulesetInstance)).ToArray();
 
                     var parsedScore = new ProcessorScoreDecoder(working).Parse(scoreInfo);
 
-                    var difficultyCalculator = rulesetInstance.CreateDifficultyCalculator(working);
+                    int modsHash = RulesetHelper.GenerateModsHash(mods, working.BeatmapInfo.Difficulty, ruleset.Value);
 
-                    Mod[] mods = score.Mods.Select(x => x.ToMod(rulesetInstance)).ToArray();
+                    if (!attributesCache.TryGetValue(modsHash, out var difficultyAttributes))
```
stanriders, line 241: Realistically one player wouldn't have more than like ~10 scores on average with the same mod combination on one map so it seems redundant to me.

PerformanceCalculatorGUI/RulesetHelper.cs
```diff
@@ -87,6 +89,74 @@ public static Ruleset GetRulesetFromLegacyID(int id)
             };
         }
 
+        /// <summary>
+        /// Generates the unique hash of mods combo that affect difficulty calculation
+        /// Needs to be updated if list of difficulty adjusting mods changes
+        /// </summary>
+        public static int GenerateModsHash(Mod[] mods, BeatmapDifficulty difficulty, RulesetInfo ruleset)
```
stanriders, line 96: I really want this to not exist - not only is it one more thing to keep updated, its also redundant and confusing to look at

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
```diff
@@ -0,0 +1,142 @@
...
+        public LazerCalculationSettings()
+        {
+            TooltipMain = "Calculation Settings";
+
+            SetIcon(new ScreenSelectionButtonIcon(FontAwesome.Solid.Cog) { IconSize = new Vector2(70) });
+        }
+
+        public bool ShouldBeFiltered(ScoreInfo score)
+        {
+            if (score.Mods.Any(h => h is OsuModAutopilot))
```
stanriders, line 47: Why?..

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -6,22 +6,27 @@
 using System.Linq;
```
stanriders, line 6: Please move database calculation into a separate screen

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -282,11 +383,213 @@ private void calculateProfile(string username)
...
+                var lazerPath = configManager.GetBindable<string>(Settings.LazerFolderPath).Value;
+
+                if (lazerPath == string.Empty)
+                {
+                    notificationDisplay.Display(new Notification("Please set-up path to lazer database folder in GUI settings"));
+                    return;
+                }
+
+                var storage = gameHost.GetStorage(lazerPath);
+                var realmAccess = new RealmAccess(storage, @"client.realm");
```
stanriders, line 447: Tools should never use lazer's database directly - please copy to a local directory and use that instead

stanriders, line 447: `RealmAccess` definitely should not be created here - having it in a `Task` is bound to have threading-related issues

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -282,11 +383,213 @@ private void calculateProfile(string username)
...
+                        }
+
+                        var perfAttributes = await performanceCalculator?.CalculateAsync(score, difficultyAttributes, token)!;
+
+                        score.PP = perfAttributes?.Total ?? 0.0;
+
+                        currentScoresCount++;
+
+                        // Sanity check for aspire maps till slider fix won't get merged
+                        if (difficultyAttributes.StarRating > 14 && !score.BeatmapInfo.Status.GrantsPerformancePoints())
```
stanriders, line 493: Please don't

PerformanceCalculatorGUI/Components/ExtendedProfileScore.cs
```diff
@@ -1,423 +1,115 @@
...
-using osu.Game.Rulesets.Scoring;
-using osu.Game.Rulesets.UI;
-using osu.Game.Utils;
-using osuTK;
-using PerformanceCalculatorGUI.Components.TextBoxes;
 
 namespace PerformanceCalculatorGUI.Components
 {
-    public class ExtendedScore
+    public class ExtendedProfileScore : ProfileScore
```
stanriders, line 16: What's the reasoning for this split?

stanriders (changes requested):

minisbett:
> > Selecting maps from lazer instead of having to bother with .osu files? yes please merge this that'd be sooo useful
> 
> This is not what it does. It's only allowing you to calculate already existing scores from your lazer scores database. You can also just use beatmap ids instead of .osu files already

Oh, it sounded like there's a proper selection for a beatmap with a search function instead of having to find the beatmap id, but seems like i misunderstood that

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
minisbett: Then please add the least add buttons to copy lazers data into a local copy and be able to sync it

stanriders: Please don't, it should be syncronized automatically at the start of profile calculation.

minisbett: oh thats an option too, but why should it be cloned instead of accessing Lazer directly in the first place? Issues if Lazer is running at the same time?

PerformanceCalculatorGUI/Screens/BeatmapLeaderboardScreen.cs
Givikap120: Caching reduced calculation time in 10 times for me. 10 times performance difference, especially in case of 15 seconds vs 3 minutes is definitely not redundant.

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: > Tools should never use lazer's database directly - please copy to a local directory and use that instead

copying 30+ gigs of data? sounds very bad

PerformanceCalculatorGUI/Components/ExtendedProfileScore.cs
Givikap120: Because I need to have Profile Score class that display only pp (as it doesn't have delta). And Extended Profile Score (child of Profile Score) also displays delta.

Givikap120:
> > > Selecting maps from lazer instead of having to bother with .osu files? yes please merge this that'd be sooo useful
> > 
> > 
> > This is not what it does. It's only allowing you to calculate already existing scores from your lazer scores database. You can also just use beatmap ids instead of .osu files already
> 
> Oh, it sounded like there's a proper selection for a beatmap with a search function instead of having to find the beatmap id, but seems like i misunderstood that

If Realm Access of this PR is merged - then making proper selection for a beatmap is 10 times easier as it's the main problem that's stopping search in lazer db.

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
stanriders: Your `client.realm` is _30 gigabytes_? That doesn't sound right at all

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: Many players played aspire maps. And they're very broken because of bad slider aim calc formula. That makes "calculation on unranked maps" option to be pretty inaccurate, as first positions will be taken by aspire maps.

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
```diff
@@ -0,0 +1,142 @@
...
+using osu.Framework.Input.Events;
+using osu.Game.Overlays.Toolbar;
+using osu.Framework.Bindables;
+using osu.Game.Scoring;
+using osu.Game.Beatmaps;
+using osu.Game.Rulesets.Osu.Mods;
+
+namespace PerformanceCalculatorGUI.Components
+{
+    public partial class LazerCalculationSettings : ToolbarButton, IHasPopover
```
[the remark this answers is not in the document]
Givikap120: Because I needed to make "options" button and ToolbarButton class was the easiest way to do it.

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
Givikap120: Autopilot is broken
https://github.com/ppy/osu/pull/21211

stanriders: I don't see how it's broken from the PR you linked. Excluding _just one mod_ doesn't make sense to me

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
stanriders: This is not a player-focused tool, excluding maps because they're "broken" is making it worse to use

PerformanceCalculatorGUI/RulesetHelper.cs
Givikap120: Calculating billion of scores without caching is pretty much impossible
Having to update one function once new mod affect Star Rating (happens once in a decade) is much better than having to wait for 30 minutes till your recalc will finish

PerformanceCalculatorGUI/Screens/BeatmapLeaderboardScreen.cs
stanriders: Calculating my own database (4410 scores) took 3 minutes 30 seconds with caching and 2 minutes 8 seconds without. I'm not saying that caching is making it slower (its probably just filesystem being faster when accessing the same files again), but it's definitely isn't making it faster

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: It initially was "RealmScreen", but first - it have big amount of code duplication, second - it's much more convenient to use as a switch for profile calculation (the same way as simulation screen have options for import by ID and import by .osu file). 
But if it's important to have separate screens for this - I will revert the merge of 2 screens.

PerformanceCalculatorGUI/Screens/BeatmapLeaderboardScreen.cs
stanriders: After running it couple more times with and without caching it settled at about the same time per calculation with non-cached being ~15 seconds slower which I think is negligible considering you already have to wait a couple of minutes per run

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: You said lazer database, it's 30 gigs for me. client.realm is 50M

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
Givikap120: Just it's the most blatantly overweight as it gives as much pp as NM. Maybe there should be "broken mods" list.
Because AP and DT RA being in the same category "unranked" mods is very bad.

PerformanceCalculatorGUI/Screens/BeatmapLeaderboardScreen.cs
Givikap120: > After running it couple more times with and without caching it settled at about the same time per calculation with non-cached being ~15 seconds slower which I think is negligible considering you already have to wait a couple of minutes per run

And now try to disable "enable scorev1-ning" switch and then re-test

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
stanriders: Lazer database is `client.realm`

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
stanriders: There should not be any "broken mods list" to begin with. https://github.com/ppy/osu-tools/pull/213#discussion_r1739653096

Givikap120: ok

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: ok, fixed

PerformanceCalculatorGUI/Screens/BeatmapLeaderboardScreen.cs
Givikap120: > After running it couple more times with and without caching it settled at about the same time per calculation with non-cached being ~15 seconds slower which I think is negligible considering you already have to wait a couple of minutes per run

I don't know how you tested it, but you definitely messed something up, because you can't get this low speed-up from just simply accessing array of attributes instead of calculating it each time
https://www.youtube.com/watch?v=KEYa7Y-UJCw
cached: 0:36 - 3:17 = 161s = 2m 41s
not cached: 4:24 - 11:09 = 405s = 6m 45s
the difference is 2.5 times (not 10, but I haven't tested it before, I just noticed that it was giga slow without caching)

PerformanceCalculatorGUI/Components/ProfileScore.cs
```diff
@@ -0,0 +1,380 @@
...
+            {
+                Title = score.BeatmapInfo?.Metadata.Title ?? "Error Title",
+                TitleUnicode = score.BeatmapInfo?.Metadata.TitleUnicode ?? "Error Title",
+                Artist = score.BeatmapInfo?.Metadata.Artist ?? "Error Artist",
+                ArtistUnicode = score.BeatmapInfo?.Metadata.ArtistUnicode ?? "Error Artist",
+            };
+            APIBeatmap dummyBeatmap = new APIBeatmap
+            {
+                OnlineID = score.BeatmapInfo?.OnlineID ?? 0,
+                DifficultyName = score.BeatmapInfo?.DifficultyName ?? "Error Difficulty",
```
smoogipoo, line 65: "Unknown" maybe?

PerformanceCalculatorGUI/RulesetHelper.cs
smoogipoo: This cannot exist. You cannot combine 50 million different flags like this into a hash function.

The most that should be done is akin to https://github.com/ppy/osu/blob/62f737d8de0944d5b278d4ab72844af6dbc67ace/osu.Game/Beatmaps/BeatmapDifficultyCache.cs#L295-L305

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
```diff
@@ -0,0 +1,139 @@
...
+                                Current = { BindTarget = bindables[2] }
+                            },
+                            new OsuCheckbox
+                            {
+                                LabelText = "Calculate Unranked Mods, Autopilot is excluded regardless",
+                                Current = { BindTarget = bindables[3] }
+                            },
+                            new OsuCheckbox
+                            {
+                                LabelText = "Enable Scorev1 score overwrite for legacy scores",
```
smoogipoo, line 130: What does this even mean?

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -282,11 +383,209 @@ private void calculateProfile(string username)
...
+            calculationCancellatonToken = new CancellationTokenSource();
+            var token = calculationCancellatonToken.Token;
+
+            Task.Run(async () =>
+            {
+                Schedule(() => loadingLayer.Text.Value = "Getting user data...");
+
+                var player = await apiManager.GetJsonFromApi<APIUser>($"users/{username}/{ruleset.Value.ShortName}");
+
+                currentUser = [player.Username, .. player.PreviousUsernames, player.Id.ToString()];
```
smoogipoo, line 416: I have never seen this syntax ever before, please don't use it.

smoogipoo (changes requested):
I honestly don't know if this even belongs in this repo... I'm doing a very broad review because I trust StanR with general code quality and what is/isn't acceptable.

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
smoogipoo: It should be fine as long as it's constrained to a single thread. If you're passing the realm or objects retrieved within it that are _not_ `.Detach()`ed between threads, then there'll be an issue.

smoogipoo: Either that or maybe have the user select the client.db?

Furthermore, I wonder if you'd be better off splitting into two processes:
1. Select a db file, copy it locally, recalculate scores given whatever parameters, _write_ values back to the DB.
2. Select a db file, display scores.

You could implement (1) as a CLI command that writes the DB to an output path.
```
dotnet run -- recalc-lazer-db /path/to/input/client.ream /path/to/output/client.realm [--options]
```

stanriders: It already crashes on someone's linux machine because of cross-thread access, I think it'd just be safer to use it through safeguards instead of hoping that TPL wouldn't do something funny

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
Givikap120: In stable for each unique mod combination only score with the highest scorev1 is calculated. This switch is turning this mechanism on and off.

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
```diff
@@ -0,0 +1,139 @@
...
+            this.ShowPopover();
+            return base.OnClick(e);
+        }
+    }
+
+    public partial class LazerCalculationSettingsPopover : OsuPopover
+    {
+        private readonly Bindable<bool>[] bindables;
+
+        public LazerCalculationSettingsPopover(Bindable<bool>[] bindables)
```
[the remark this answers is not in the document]
Givikap120: fixed

PerformanceCalculatorGUI/Components/ProfileScore.cs
Givikap120: fixed

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
smoogipoo: What does that have to do with lazer? It doesn't sound like you've properly thought through what this PR is set to achieve and are just throwing flags around the place in the name of optimisation...

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
bdach: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions

> Collection expressions introduce a new terse syntax to create common collection values. Inlining other collections into these values is possible using a spread element `..e`.

I found spread operators pretty useful to work with in other languages so I'd be a little bummed to see them banished.

smoogipoo: Maybe it's the usage that's confusing here, because it's assigning an array to something called "currentUser". If it were "currentUserNames" it'd probably be more understandable.

I'll trade you this for literally every other feature I've pushed over the last couple of years but get strong pushback every time because it's some newspangled wankjizz... No doubt I'm going to be the only one of us core devs who sees this feature and questions it for its one esoteric use. :(

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: Is there a way to make copied client.realm to contain ONLY replay files
not only this will be more convenient - it will be also faster

smoogipoo: Faster how? You said it's a 50MB file? That takes 2 seconds on a 2000s era HDD, and is practically instantaneous on anything faster.

I suggest you don't go pre-optimising.

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: So I'm gonna just rename `currentUser`?

PerformanceCalculatorGUI/Components/LazerCalculationSettings.cs
Givikap120: this is not about optimization
it's about giving user a choice between "accurate simulation" and "what if all my scores was weighted fairly"
for now it's not making big difference, but when combining with CSR - the difference in total pp can be quite big

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
Givikap120: This is not about copying
The first 28 seconds of 3 minutes calculation was spent to collect all scores from the database, what is quite big portion of the time
Source:
https://www.youtube.com/watch?v=KEYa7Y-UJCw

smoogipoo: That is a problem with the structure of your code, because you're using `Detach()`. You should restructure so you don't have to do that, such as using the `Live<>` class or otherwise.

PR #215, opened 2024-09-10
Improve object inspection visibility

stanriders:
Makes previous hitobject slightly transparent to show that its expired and adds proper fade out animations instead of abruptly deleting the object

image


peppy:
Just gonna merge this one as it seems fine.

peppy (approved):

PR #218, opened 2024-10-08
Disable tablet input handler on start

stanriders:
Some people are having native driver crashes because of OTD being enabled by default

### DIFF
--- a/PerformanceCalculatorGUI/PerformanceCalculatorGame.cs
+++ b/PerformanceCalculatorGUI/PerformanceCalculatorGame.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
 using osu.Framework.Graphics;
+using osu.Framework.Input.Handlers.Tablet;
 using osu.Framework.Platform;
 using osu.Game;
 using osu.Game.Graphics.Cursor;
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):

peppy:
Wouldn't this imply the same thing would happen on lazer? Or is there something specifically different here?

stanriders:
> Wouldn't this imply the same thing would happen on lazer? Or is there something specifically different here?
> 
> What do the crashes look like?

That would happen on lazer too I believe, but there you have a toggle to disable the tablet driver and it wont interfere anymore.

Lazer's tablet driver crashes xp-pen's tablet driver. It was reported by a user, and I didn't really ask for additional details since disabling tablet input in tools is definitely something we'd want to do anyway (because it'd interfere if for example user tries to play the game at the same time as tools are open, or just manufacturers drivers that are open in background being unhappy)

peppy:
SO just one user, or multiple? "Some people" implies more than one.

COuld you please include the raw reports of this issue here so we can actually look to fix it if required? I don't think this kind of change should be made to an auxiliary project like this for one user.

stanriders:
I didn't consider this important enough, as it's expected that having two drivers for one device isn't a good idea and might lead to one of them breaking, and again disabling tablet input in tools is something that should be done anyway.

I'll ask for more details about the crash

BunnyTub:
I did have this issue with my Wacom tablet before, but stopped being an issue when I destroyed the drivers and switched to osu!'s OpenTabletDriver, then a bit after that, switched to the OpenTabletDriver daemon (with osu!'s OpenTabletDriver disabled).

PR #219, opened 2024-10-13
Add option to input large tick misses to simulate screen

minisbett:
Adds a text field for specifying the `HitResult.LargeTickMiss` statistic of the `ScoreInfo` simulated on the simulation screen.
This hit result is currently not considered in PP calculations but will be soon via ppy/osu#27691.

image

### DIFF
--- a/PerformanceCalculatorGUI/RulesetHelper.cs
+++ b/PerformanceCalculatorGUI/RulesetHelper.cs
@@ -103,11 +103,11 @@ namespace PerformanceCalculatorGUI
             return (int)Math.Round(1000000 * scoreMultiplier);
         }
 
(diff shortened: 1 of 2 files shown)


smoogipoo:
This should probably detect what judgements are generally valid for rulesets and accept arbitrary inputs, but let's start with this because there's not really good support for that right now... We have a very primitive [`GetValidHitResults`](https://github.com/ppy/osu/blob/2de19558923160116a0139a955fa431008ef5d89/osu.Game.Rulesets.Mania/ManiaRuleset.cs#L383-L396) but that's not even covering the [`ComboBreak`](https://github.com/ppy/osu/blob/2de19558923160116a0139a955fa431008ef5d89/osu.Game.Rulesets.Mania/Judgements/HoldNoteBodyJudgement.cs#L11) result for mania so it's very imperfect.

smoogipoo (approved):

PR #220, opened 2024-10-13
change default beatmap import type of simulation screen to ID

minisbett:
After asking around, as I expected most people rather use the ID import type instead of specifying a local file. Saves a click, especially when restarting alot.


smoogipoo (approved):
Sounds reasonable to me.

PR #223, opened 2024-10-18
Move `SimulateCommand` option implementations to base class

minisbett:
Moves the option attributes of all parameters of the simulate command where every ruleset has the same implementations to the base class. The reduces the amount of boilerplate code in each simulate command class.

The only parameters excluded are:
- Mehs/Goods, since Catch renames them to droplets and Taiko does not have Mehs
- Combo/ComboPercent, as Mania does not have combo


minisbett:
Roughly tested, worth it giving two more eyes a quick check on it

smoogipoo (approved):

Issue #226, opened 2024-10-23
Incorrect accuracy calculation for lazer


MaxOhn:
Haven't checked for other modes but at least for osu!standard the accuracy calculation is no longer correct for lazer since it does not account for slider tick and slider end hits.

https://github.com/ppy/osu-tools/blob/156867cba3fa57bbc0f7532f1c2d186063c83fde/PerformanceCalculator/Simulate/OsuSimulateCommand.cs#L135-L144

This is important because after generating the hitresults it calculates the accuracy based on those and then passes the calculated accuracy to the performance calculation.

https://github.com/ppy/osu-tools/blob/156867cba3fa57bbc0f7532f1c2d186063c83fde/PerformanceCalculator/Simulate/SimulateCommand.cs#L69-L76

The performance calculation then may or may not make use of that accuracy. In case of osu!standard, [it does make use of it](https://github.com/ppy/osu/blob/c1453cfd0cf858579e6d52b212598f8233124ad8/osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs#L54).

I imagine this will be slightly awkward to fix. Perhaps a new boolean flag `--lazer` that will be passed to accuracy calculation to decide whether new or old accuracy calculation should be used?

MaxOhn:
On a somewhat related note:

https://github.com/ppy/osu-tools/blob/156867cba3fa57bbc0f7532f1c2d186063c83fde/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs#L42-L43

Adding the hold note count here should only be done for lazer but not stable. A new `--lazer` flag could be used here too.

minisbett:
> Copied



> On a somewhat related note:
> 
> https://github.com/ppy/osu-tools/blob/156867cba3fa57bbc0f7532f1c2d186063c83fde/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs#L42-L43
> 
> Adding the hold note count here should only be done for lazer but not stable. A new `--lazer` flag could be used here too.

Is this a difference between CL and non-CL or stable vs. lazer? It'd be ideal if flags specific to Lazer can be avoided and rather specified through mods.

Aside, this is a known issue and I've wanted to fix it, but sort-of gave up for now because the calculations involve having to pass around beatmap objects for maximum statistics, so at least for me it was not a quick thing to do in a pleasing way.

PR #227, opened 2024-10-23
Refactor `GetMaxCombo`

MaxOhn:
Arguably for mania it should stay on zero but I'd say this should be fine.

Notably, if max combo calculation in lazer ever differs from stable for any mode, this will require better handling. As far as I know they currently coincide.

### DIFF
--- a/PerformanceCalculator/Simulate/SimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/SimulateCommand.cs
@@ -65,7 +65,7 @@ namespace PerformanceCalculator.Simulate
             var mods = ParseMods(ruleset, Mods, ModOptions);
             var beatmap = workingBeatmap.GetPlayableBeatmap(ruleset.RulesetInfo, mods);
 
-            var beatmapMaxCombo = GetMaxCombo(beatmap);
+            var beatmapMaxCombo = beatmap.GetMaxCombo();
             var statistics = GenerateHitResults(Accuracy / 100, beatmap, Misses, Mehs, Goods);
             var scoreInfo = new ScoreInfo(beatmap.BeatmapInfo, ruleset.RulesetInfo)
             {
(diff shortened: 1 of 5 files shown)


PerformanceCalculator/Simulate/SimulateCommand.cs
```diff
@@ -83,7 +83,7 @@ public override void Execute()
             OutputPerformance(scoreInfo, performanceAttributes, difficultyAttributes);
         }
 
-        protected abstract int GetMaxCombo(IBeatmap beatmap);
+        protected static int GetMaxCombo(IBeatmap beatmap) => beatmap.GetMaxCombo();
```
stanriders, line 86: At this point you might as well remove this completely and replace all usages with `beatmap.GetMaxCombo()`

MaxOhn: Was my plan at first but noticed that the hitresult calculation for catch and taiko make use of this method so I figured it would be better to keep to make sure it stays consistent.

MaxOhn:
After checking some mania leaderboards it seems that max combo does differ between lazer and stable. If this PR goes through as-is, that means the shown value will be incorrect when intending to calculate stable attributes. The actual calculated attributes won't be affected though considering that their calculation in mania (currently) ignores combo.

Regardless, the shown value will be closer to the actual combo than just `0` 😄 An option to distinguish lazer from stable attribute calculation as warranted in #226 could help fixing this entirely.

minisbett:
In my opinion, the correct implementation here would be re-adding the `GetMaxCombo` method as a `virtual` method with a default implementation of calling `beatmap.GetMaxCombo()`. Then, for Mania it can be overriden to include the differences between stable and lazer.

This was a refactor I wanted to do some days ago but I got caught up in other stuff and haven't found time.

MaxOhn:
I removed the method entirely now 👌 

Given that the max combo is in somewhat of a limbo until the stable-lazer difference can be handled properly in osu-tools, maybe it's better to hold off on this?

minisbett:
> I removed the method entirely now 👌 
> 
> Given that the max combo is in somewhat of a limbo until the stable-lazer difference can be handled properly in osu-tools, maybe it's better to hold off on this?

I don't think it will be handled differently since there isn't really a need to. Mania has that combo difference as (afaik?) the only ruleset, and combo does not matter in PP there so its fine to go with Lazer combo, it's pretty much just display.

minisbett (approved):
Replacing the direct calculation of maximum combo with the `IBeatmap.GetMaxCombo()` call seems to match 1:1 (tested) (except for differences between stable and lazer on mania), therefore this seems like a good change.

Take my meaningless approval.

stanriders (approved):
I'd say we'd want to rely on lazer's methods as much as possible anyway

PR #229, opened 2024-11-01
Fix standard ruleset accuracy calculation on Lazer

### LINKED ISSUE #226: Incorrect accuracy calculation for lazer

minisbett:
Resolves #226 

Adds large tick hits and slider tail hits to the accuracy calculation. For that, the `IBeatmap` is now passed into the `GetAccuracy` method, although unused on any ruleset except standard.

### DIFF
--- a/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
@@ -65,7 +65,7 @@ namespace PerformanceCalculator.Simulate
             };
         }
 
-        protected override double GetAccuracy(Dictionary<HitResult, int> statistics)
+        protected override double GetAccuracy(IBeatmap beatmap, Dictionary<HitResult, int> statistics)
         {
             double hits = statistics[HitResult.Great] + statistics[HitResult.LargeTickHit] + statistics[HitResult.SmallTickHit];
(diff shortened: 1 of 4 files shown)


smoogipoo (approved):
Looks okay

PR #230, opened 2024-11-15
Fix accuracy formula being wrong for lazer scores in GUI

Givikap120:
It used stable formula before.
Now it will use lazer formula automatically if lazer info is available.

### DIFF
--- a/PerformanceCalculatorGUI/RulesetHelper.cs
+++ b/PerformanceCalculatorGUI/RulesetHelper.cs
@@ -21,7 +21,6 @@ using osu.Game.Rulesets.Taiko;
 using osu.Game.Rulesets.Taiko.Objects;
 using osu.Game.Skinning;
 using osu.Game.Utils;
-using osuTK;
 
 namespace PerformanceCalculatorGUI
 {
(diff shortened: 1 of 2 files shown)


PerformanceCalculatorGUI/RulesetHelper.cs
```diff
@@ -293,27 +297,47 @@ private static Dictionary<HitResult, int> generateManiaHitResults(double accurac
...
+                max += 3 * countSliders;
+            }
+
+            if (statistics.ContainsKey(HitResult.LargeTickMiss))
+            {
+                var countLargeTicks = beatmap.HitObjects.Sum(obj => obj.NestedHitObjects.Count(x => x is SliderTick or SliderRepeat));
+                var countLargeTickHit = countLargeTicks - statistics[HitResult.LargeTickMiss];
+
+                total += 0.6 * countLargeTickHit;
+                max += 0.6 * countLargeTicks;
```
smoogipoo, line 337: By the way, these need to not use `6`, `2`, `3`, `0.6`...

It should be creating a `ScoreProcessor` and calling `GetBaseScoreForResult()`.

smoogipoo (approved):

PR #231, opened 2024-11-15
Refactor HitResult generation to make ruleset-specific behavior more explicit

Givikap120:
Before you had function `GenerateHitResults(double accuracy, IBeatmap beatmap, int countMiss, int? countMeh, int? countGood)`, where you pass `countMiss`, `countMeh` and `countGood`.
So by function signature you may assume that it's using those 3 judgements to generate hitresults.
In reality - osu also used `largeTickMisses` and `sliderTailMisses` from class itself, mania used `oks` and `greats` from class, when taiko just ignored `countMeh` because it doesn't exist in taiko. What is quite confusing and unintuitive.

This PR removes those params from basic `GenerateHitResults`, so it would be intuitive that all values are taken from attributes of the class.
And then this function delegates calculation to ruleset-specific static function that takes correct parameters.

Prerequisite for https://github.com/ppy/osu-tools/pull/232

### DIFF
--- a/PerformanceCalculator/Simulate/SimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/SimulateCommand.cs
@@ -63,7 +63,7 @@ namespace PerformanceCalculator.Simulate
 
             var workingBeatmap = ProcessorWorkingBeatmap.FromFileOrId(Beatmap);
             var mods = ParseMods(ruleset, Mods, ModOptions);
-            var beatmap = workingBeatmap.GetPlayableBeatmap(ruleset.RulesetInfo);
+            var beatmap = workingBeatmap.GetPlayableBeatmap(ruleset.RulesetInfo, mods);
 
             var beatmapMaxCombo = beatmap.GetMaxCombo();
             var statistics = GenerateHitResults(beatmap);
--- a/PerformanceCalculator/Simulate/OsuSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/OsuSimulateCommand.cs
@@ -45,7 +45,7 @@ namespace PerformanceCalculator.Simulate
 
         protected override Dictionary<HitResult, int> GenerateHitResults(IBeatmap beatmap) => generateHitResults(beatmap, Accuracy / 100, Misses, Mehs, Goods, largeTickMisses, sliderTailMisses);
 
-        private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, double accuracy, int countMiss, int? countMeh, int? countGood, int? countLargeTickMisses, int? countSliderTailMisses)
+        private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, double accuracy, int countMiss, int? countMeh, int? countGood, int countLargeTickMisses, int countSliderTailMisses)
         {
             int countGreat;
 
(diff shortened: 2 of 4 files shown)


minisbett:
This PR, similar to #232, seems to complicate code at no gain. Nothing is wrong with the current state, and imo splitting this up is unnecessary

Givikap120:
> This PR, similar to #232, seems to complicate code at no gain. Nothing is wrong with the current state, and imo splitting this up is unnecessary

I explain the gain in detail in PRs description. Please respond more precisely about what is trying to be addressed instead of "PR is bad".

PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
```diff
@@ -66,28 +68,28 @@ protected override Dictionary<HitResult, int> GenerateHitResults(double accuracy
...
             // Everything else is a meh, as initially assumed.
             countMeh = remainingHits;
 
             return new Dictionary<HitResult, int>
             {
                 { HitResult.Perfect, perfects },
-                { HitResult.Great, countGreat },
-                { HitResult.Ok, countOk },
+                { HitResult.Great, greats },
+                { HitResult.Ok, oks },
```
minisbett, line 92: Why rename those to greats and oks? With `countGreat` and `countOks` they fit into the naming of the other variables

Givikap120: because it was initially named in a wrong way

look at before
```c#
return new Dictionary<HitResult, int>
                {
                    [HitResult.Perfect] = countPerfect,
                    [HitResult.Great] = greats ?? 0,
                    [HitResult.Good] = countGood ?? 0,
                    [HitResult.Ok] = oks ?? 0,
                    [HitResult.Meh] = countMeh ?? 0,
                    [HitResult.Miss] = countMiss
                };
```

and after
```c#
return new Dictionary<HitResult, int>
                {
                    [HitResult.Perfect] = countPerfect,
                    [HitResult.Great] = countGreat ?? 0,
                    [HitResult.Good] = countGood ?? 0,
                    [HitResult.Ok] = countOk ?? 0,
                    [HitResult.Meh] = countMeh ?? 0,
                    [HitResult.Miss] = countMiss
                };
```

minisbett: But why are you changing them back to the old names?

Givikap120: I don't know what this means
I just swapped countOk with oks and countGreat with greats to be consistent with context

minisbett: In a previous PR i sawpped `oks` and `greats` with `countOk` and `countGreat` in order to match the name of the other variables

Givikap120: look on the function signature
`private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, double accuracy, int countMiss, int? countMeh, int? countOk, int? countGood, int? countGreat)`

do in your opinion I should change it to 
`private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, double accuracy, int countMiss, int? countMeh, int? oks, int? countGood, int? greats)`

looks like unreasonable and inconsistent arguments names

minisbett: ohhh, I didn't see that this is an issue with clashing names.

PerformanceCalculator/Simulate/SimulateCommand.cs
```diff
@@ -63,10 +63,10 @@ public override void Execute()
 
             var workingBeatmap = ProcessorWorkingBeatmap.FromFileOrId(Beatmap);
             var mods = ParseMods(ruleset, Mods, ModOptions);
-            var beatmap = workingBeatmap.GetPlayableBeatmap(ruleset.RulesetInfo, mods);
+            var beatmap = workingBeatmap.GetPlayableBeatmap(ruleset.RulesetInfo);
```
minisbett, line 66: Why was the mods parameter removed here? I believe there's mods that alter the beatmap in a way it is relevant here, eg. maybe conversion mods?

Givikap120: oops, this is accidental

smoogipoo:
@Givikap120 you're going to need to fix this PR.

smoogipoo:
Will also await on further refactoring as per https://discord.com/channels/188630481301012481/380598781432823815/1307180578906312724 before merging.

minisbett:
> Will also await on further refactoring as per https://discord.com/channels/188630481301012481/380598781432823815/1307180578906312724 before merging.

This PR is not the one requiring refactoring I believe, this PR should be fine as-is. The accuracy calculation PR does though.

Givikap120:
@smoogipoo I fixed merge conflicts here

PR #232, opened 2024-11-15
Account for CL mod when generating HitResults

Givikap120:
Depends on #231 

Adds mods parameter `GenerateHitResults` so you can generate correct hitresults for CL scores.

### DIFF
--- a/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
@@ -35,7 +35,9 @@ namespace PerformanceCalculator.Simulate
 
         public override Ruleset Ruleset => new CatchRuleset();
 
-        protected override Dictionary<HitResult, int> GenerateHitResults(double accuracy, IBeatmap beatmap, int countMiss, int? countMeh, int? countGood)
+        protected override Dictionary<HitResult, int> GenerateHitResults(IBeatmap beatmap) => generateHitResults(beatmap, Accuracy / 100, Misses, Mehs, Goods);
+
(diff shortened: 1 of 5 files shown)


minisbett:
This PR seems to make things more complicated at no gain. The accuracy calculation already works fine with CL and non-CL, even when exclusively generating non-CL hit statistics and using the Lazer accuracy formula.

Givikap120:
> This PR seems to make things more complicated at no gain. The accuracy calculation already works fine with CL and non-CL, even when exclusively generating non-CL hit statistics and using the Lazer accuracy formula.

It's literally incorrect for CL scores

minisbett:
Oh I think you're right. Can you maybe close this PR and include these changes in #231 instead? It's annoying to review it this way

Givikap120:
> Oh I think you're right. Can you maybe close #231? I think it makes more sense to have all of this in a single PR, easier to review

I split it exactly for reason of being easier to review.
Because adding `mods` param to the function without refactor makes confusing way of calculating hitresults even more confusing.

minisbett:
> > Oh I think you're right. Can you maybe close #231? I think it makes more sense to have all of this in a single PR, easier to review
> 
> I split it exactly for reason of being easier to review. Because adding `mods` param to the function without refactor makes confusing way of calculating hitresults even more confusing.

ok yeah fair, after looking at the code for longer maybe

smoogipoo (approved):

smoogipoo:
Ah, I didn't notice the dependency here...

minisbett:
Wasn't it discussed in #difficulty-osu that the complicated accuracy calculation here can be simplified here? https://discord.com/channels/188630481301012481/380598781432823815/1307180578906312724

minisbett:
So maybe revert this PR until #231 is merged and a more proper accuracy calculation implemented?

PR #233, opened 2024-11-20
Apply adjustments for struct attributes

minisbett:
Prereqs:

[ ] https://github.com/ppy/osu/pull/30727
[ ] nuget package updates

### DIFF
--- a/PerformanceCalculator/LegacyHelper.cs
+++ b/PerformanceCalculator/LegacyHelper.cs
@@ -58,26 +58,5 @@ namespace PerformanceCalculator
                     return "mania";
             }
         }
-
-        public static DifficultyAttributes CreateDifficultyAttributes(int legacyId)
-        {
-            switch (legacyId)
-            {
-                case 0:
-                    return new OsuDifficultyAttributes();
-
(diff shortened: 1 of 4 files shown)


PerformanceCalculator/LegacyHelper.cs
```diff
@@ -58,26 +58,5 @@ public static string GetRulesetShortNameFromId(int id)
                     return "mania";
             }
         }
-
```
minisbett, line 61: I removed this method as it is unused

PerformanceCalculator/LegacyHelper.cs
```diff
@@ -58,26 +58,5 @@ public static string GetRulesetShortNameFromId(int id)
...
-                case 2:
-                    return new CatchDifficultyAttributes();
-
-                case 3:
-                    return new ManiaDifficultyAttributes();
-
-                default:
-                    throw new ArgumentException($"Invalid ruleset ID: {legacyId}", nameof(legacyId));
-            }
-        }
```
minisbett, line 81: removed this method as it is unused

PR #247, opened 2025-01-31
Implement support for viewing multiple profiles at once

buyaspacecube:
These changes let you input a list of usernames and view their combined plays on one profile page, with each play showing the avatar of the user that set it. An option is included to only display the best play on each individual beatmap


stanriders (changes requested):
This shouldn't be a separate screen, instead extend the already existing `ProfileScreen`

PerformanceCalculatorGUI/Components/ExtendedCombinedProfileScore.cs
```diff
@@ -0,0 +1,53 @@
...
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Game.Online.API.Requests.Responses;
+using osu.Game.Rulesets;
+using osuTK;
+using osu.Game.Users.Drawables;
+
+namespace PerformanceCalculatorGUI.Components
+{
+    public partial class ExtendedCombinedProfileScore : ExtendedProfileScore
```
stanriders, line 13: Add a `ShowAvatar` setting to the `ExtendedProfileScore` and draw avatar there instead of making a new class

buyaspacecube: since the APIUser needs to be passed in to get the avatar, would it make sense to add a nullable APIUser to `ExtendedProfileScore` and display the avatar if it isn't null?

buyaspacecube: oh i had no idea thank you

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -236,22 +257,11 @@ private void calculateProfile(string username)
...
-                var player = await apiManager.GetJsonFromApi<APIUser>($"users/{username}/{ruleset.Value.ShortName}");
-
-                currentUser = player.Username;
-
                 Schedule(() =>
                 {
                     if (userPanel != null)
                         userPanelContainer.Remove(userPanel, true);
 
-                    userPanelContainer.Add(userPanel = new UserCard(player)
```
stanriders, line 250: This can still happen here, just under the `if (calculatingSingleProfile)` check

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -268,58 +278,100 @@ private void calculateProfile(string username)
...
 
                 var plays = new List<ExtendedScore>();
-
+                var players = new List<APIUser>();
                 var rulesetInstance = ruleset.Value.CreateInstance();
 
-                Schedule(() => loadingLayer.Text.Value = $"Calculating {player.Username} top scores...");
+                foreach (string username in usernameArray)
+                {
+                    Schedule(() => loadingLayer.Text.Value = "Getting user data...");
```
stanriders, line 286: Might as well add _which_ user

stanriders (changes requested):
Haven't tested yet, just a quick code review

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
buyaspacecube: it's easier down there since `player` isnt defined up here anymore

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -48,7 +49,7 @@ public partial class ProfileScreen : PerformanceCalculatorScreen
         private Container userPanelContainer;
         private UserCard userPanel;
 
-        private string currentUser;
+        private string currentUsers;
```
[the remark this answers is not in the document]
buyaspacecube: all done

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
buyaspacecube: good idea

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
```diff
@@ -209,20 +224,26 @@ private void load()
...
-            if (string.IsNullOrEmpty(username))
+            if (string.IsNullOrEmpty(usernames))
             {
                 usernameTextBox.FlashColour(Color4.Red, 1);
                 return;
             }
 
+            Array.Clear(currentUsers);
+
+            string[] usernameArray = usernames.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```
stanriders, line 244: I'd suggest doing splitting only when committing from the textbox, and changing `calculateProfile()` to accept an array

stanriders (changes requested):

PerformanceCalculatorGUI/Screens/ProfileScreen.cs
buyaspacecube: good idea, i've done that and also renamed the method `calculateProfiles` so it makes more sense. i can revert that if necessary

buyaspacecube:
feel free to change the placeholder names to whatever btw

stanriders (approved):

PerformanceCalculatorGUI/Components/ExtendedProfileScore.cs
```diff
@@ -111,15 +118,25 @@ private void load(RulesetStore rulesets)
...
-                Children = new Drawable[]
+                Children = new[]
                 {
+                    ShowAvatar
+                        ? new ClickableAvatar(Score.SoloScore.User, true)
+                        {
+                            Masking = true,
+                            CornerRadius = ExtendedLabelledTextBox.CORNER_RADIUS,
+                            Size = new Vector2(avatar_size),
+                            Action = () => { host.OpenUrlExternally($"https://osu.ppy.sh/u/{Score.SoloScore.User?.Id}"); }
```
tsunyoku, line 129:
```suggestion
                            Action = () => { host.OpenUrlExternally($"https://osu.ppy.sh/users/{Score.SoloScore.User?.Id}"); }
```

`/u/{id}` is a legacy URL, whilst there is a redirect from `/u` to `/users` I'd rather just avoid it altogether

tsunyoku (changes requested):
Minor nitpick

tsunyoku (approved):

PR #249, opened 2025-02-02
Loop through colours to prevent crash in `StrainVisualizer`

buyaspacecube:
Currently StrainVisualizer causes a crash when trying to display more than six skills as only six colours are included. This prevents that crash by looping through the list of colours for skills past the sixth

### DIFF
--- a/PerformanceCalculatorGUI/Components/StrainVisualizer.cs
+++ b/PerformanceCalculatorGUI/Components/StrainVisualizer.cs
@@ -201,7 +201,7 @@ namespace PerformanceCalculatorGUI.Components
                         {
                             RelativeSizeAxes = Axes.Both,
                             MaxValue = strainMaxValue,
-                            Values = strainLists[i % skillColours.Length]
+                            Values = strainLists[i]
                         }
                     }
                 });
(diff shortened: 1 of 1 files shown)


PerformanceCalculatorGUI/Components/StrainVisualizer.cs
```diff
@@ -196,12 +196,12 @@ private void addStrainBars(Skill[] skills, List<float[]> strainLists)
...
                         RelativeSizeAxes = Axes.Both,
                         Alpha = graphAlpha,
-                        Colour = skillColours[i],
+                        Colour = skillColours[i % skillColours.Length],
                         Child = new StrainBarGraph
                         {
                             RelativeSizeAxes = Axes.Both,
                             MaxValue = strainMaxValue,
-                            Values = strainLists[i]
+                            Values = strainLists[i % skillColours.Length]
```
stanriders: This isnt the colours array

stanriders (changes requested):

PerformanceCalculatorGUI/Components/StrainVisualizer.cs
buyaspacecube: noticed that immediately after making the pr, fixed

stanriders (approved):

PR #251, opened 2025-02-20
Fix compile failures

bdach:
Required to unblock https://github.com/ppy/osu/actions/runs/13430044700/job/37519884991

### DIFF
--- a/PerformanceCalculatorGUI/Components/UserCard.cs
+++ b/PerformanceCalculatorGUI/Components/UserCard.cs
@@ -62,12 +62,6 @@ namespace PerformanceCalculatorGUI.Components
             Action = () => { host.OpenUrlExternally($"https://osu.ppy.sh/u/{User.Id}"); };
         }
 
-        protected override void LoadComplete()
-        {
-            Status.UnbindAll();
-            Activity.UnbindAll();
-        }
-
         protected override Drawable CreateLayout()
         {
             var layout = new Container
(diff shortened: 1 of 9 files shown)


PerformanceCalculatorGUI/Components/UserCard.cs
```diff
@@ -163,7 +157,14 @@ protected override Drawable CreateLayout()
                                 }
                             }
                         }
-                    }
+                    },
+                    // this is like COMPLETELY screwed but is required to have this not die of death because `ExtendedUserPanel` expects `LastVisitMessage` to not be null
+                    // and that is created and assigned to by `CreateStatusMessage()`.
+                    // the card API is SHOCKING and needs to be taken behind the barn and old yeller'd five years ago.
+                    // it is NOT A GOOD IDEA to have A PROTECTED OVERRIDABLE METHOD THAT CAN MAKE READING OTHER PROTECTED MEMBERS DANGEROUS!!!!
+                    // or that A PROTECTED METHOD which you can FORGET TO CALL should be RESPONSIBLE FOR SETTING UP A PROTECTED FIELD!!!!
```
bdach, line 165: Sorry not sorry for this rant but I genuinely cannot respond to this otherwise.

peppy (approved):

PerformanceCalculatorGUI/Components/UserCard.cs
```diff
@@ -163,7 +157,14 @@ protected override Drawable CreateLayout()
...
+                    CreateStatusMessage(false).With(wtf => wtf.Alpha = 0),
```
stanriders, line 166: Honestly I think it'd be better to just not use `ExtendedUserPanel` at all here. Also, in my local fixes to this I've just overrode `Update()` with an empty method instead which disables status updates completely

PR #252, opened 2025-02-21
Fixes for mania simulate command

MrHeliX:
- Fixes total hits calculation for mania scores with CL mod (hold notes should only count as a single hit in stable)
- Implements GetAccuracy for mania simulate command (both for CL and non-CL scores)

### DIFF
--- a/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
@@ -36,12 +36,14 @@ namespace PerformanceCalculator.Simulate
 
         public override Ruleset Ruleset => new ManiaRuleset();
 
-        protected override Dictionary<HitResult, int> GenerateHitResults(IBeatmap beatmap, Mod[] mods) => generateHitResults(beatmap, Accuracy / 100, Misses, Mehs, oks, Goods, greats);
+        protected override Dictionary<HitResult, int> GenerateHitResults(IBeatmap beatmap, Mod[] mods) => generateHitResults(beatmap, mods, Accuracy / 100, Misses, Mehs, oks, Goods, greats);
 
-        private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, double accuracy, int countMiss, int? countMeh, int? countOk, int? countGood, int? countGreat)
+        private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, Mod[] mods, double accuracy, int countMiss, int? countMeh, int? countOk, int? countGood, int? countGreat)
         {
             // One judgement per normal note. Two judgements per hold note (head + tail).
-            int totalHits = beatmap.HitObjects.Count + beatmap.HitObjects.Count(ho => ho is HoldNote);
+            int totalHits = beatmap.HitObjects.Count;
+            if (!mods.Any(m => m.Acronym == "CL"))
+                totalHits += beatmap.HitObjects.Count(ho => ho is HoldNote);
 
             if (countMeh != null || countOk != null || countGood != null || countGreat != null)
             {
--- a/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
@@ -98,5 +98,22 @@ namespace PerformanceCalculator.Simulate
                 { HitResult.Miss, countMiss }
             };
         }
+
+        protected override double GetAccuracy(IBeatmap beatmap, Dictionary<HitResult, int> statistics)
+        {
+            int countPerfect = statistics[HitResult.Perfect];
+            int countGreat = statistics[HitResult.Great];
+            int countGood = statistics[HitResult.Good];
+            int countOk = statistics[HitResult.Ok];
+            int countMeh = statistics[HitResult.Meh];
+            int countMiss = statistics[HitResult.Miss];
+
+            double perfectWeight = Mods != null && Mods.Any(m => m == "CL") ? 300 : 305;
+
+            double total = perfectWeight * countPerfect + 300 * countGreat + 200 * countGood + 100 * countOk + 50 * countMeh;
+            double max = perfectWeight * (countPerfect + countGreat + countGood + countOk + countMeh + countMiss);
+
+            return total / max;
+        }
     }
 }


stanriders:
Would you mind doing the same for GUI? https://github.com/ppy/osu-tools/blob/master/PerformanceCalculatorGUI/RulesetHelper.cs

MrHeliX:
> Would you mind doing the same for GUI? https://github.com/ppy/osu-tools/blob/master/PerformanceCalculatorGUI/RulesetHelper.cs

Not too familiar with the GUI code but I think this does the trick

PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
```diff
@@ -36,12 +36,14 @@ public class ManiaSimulateCommand : SimulateCommand
...
-        protected override Dictionary<HitResult, int> GenerateHitResults(IBeatmap beatmap, Mod[] mods) => generateHitResults(beatmap, Accuracy / 100, Misses, Mehs, oks, Goods, greats);
+        protected override Dictionary<HitResult, int> GenerateHitResults(IBeatmap beatmap, Mod[] mods) => generateHitResults(beatmap, mods, Accuracy / 100, Misses, Mehs, oks, Goods, greats);
 
-        private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, double accuracy, int countMiss, int? countMeh, int? countOk, int? countGood, int? countGreat)
+        private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, Mod[] mods, double accuracy, int countMiss, int? countMeh, int? countOk, int? countGood, int? countGreat)
         {
             // One judgement per normal note. Two judgements per hold note (head + tail).
-            int totalHits = beatmap.HitObjects.Count + beatmap.HitObjects.Count(ho => ho is HoldNote);
+            int totalHits = beatmap.HitObjects.Count;
+            if (!mods.Any(m => m.Acronym == "CL"))
```
minisbett, line 45: The mod check should be done with `!mods.Any(x => x is ModClassic)`

PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
```diff
@@ -96,5 +98,22 @@ private static Dictionary<HitResult, int> generateHitResults(IBeatmap beatmap, d
...
+        protected override double GetAccuracy(IBeatmap beatmap, Dictionary<HitResult, int> statistics)
+        {
+            int countPerfect = statistics[HitResult.Perfect];
+            int countGreat = statistics[HitResult.Great];
+            int countGood = statistics[HitResult.Good];
+            int countOk = statistics[HitResult.Ok];
+            int countMeh = statistics[HitResult.Meh];
+            int countMiss = statistics[HitResult.Miss];
+
+            int perfectWeight = Mods != null && Mods.Any(m => m == "CL") ? 300 : 305;
```
minisbett, line 111: Here too

minisbett (changes requested):
~~On Lazer, even with CL it still does the Lazer way of doing hits, therefore you cannot simply check for CL here.~~
~~This is a bit troublesome as there's no way to tell apart Lazer and Stable, so I think the best that can be done is some command-line parameter:~~
```cs
[UsedImplicitly]
[Option(Template = "-S|--stable", Description = "The score is considered an osu! stable score.")]
public bool IsStable { get; }
```
~~(source: I forgot but when I refactored all this stuff I ran into this issue, I think @Givikap120 knows more about it)~~

**__Edit: I backread older PRs, and read that CL does not exist on Mania. Does that mean CL is a definitive marker for the score origining from Stable?__** And also, can't you use `GetMaxCombo` on mania to get the total hits?

PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
MrHeliX: The null check was added precisely because it crashed when calculating a nomod score

MrHeliX:
> ~On Lazer, even with CL it still does the Lazer way of doing hits, therefore you cannot simply check for CL here.~ ~This is a bit troublesome as there's no way to tell apart Lazer and Stable, so I think the best that can be done is some command-line parameter:~
> 
> ```cs
> [UsedImplicitly]
> [Option(Template = "-S|--stable", Description = "The score is considered an osu! stable score.")]
> public bool IsStable { get; }
> ```
> 
> ~(source: I forgot but when I refactored all this stuff I ran into this issue, I think @Givikap120 knows more about it)~
> 
> **Edit: I backread older PRs, and read that CL does not exist on Mania. Does that mean CL is a definitive marker for the score origining from Stable?** And also, can't you use `GetMaxCombo` on mania to get the total hits?

Honestly I am not too involved with mania but from what I've learnt today it seems to be all over the place. There are currently three different ways of calculating accuracy: the stable way (with CL), the Lazer way (without CL) and the website way (with judgements according to CL logic but accuracy calculation according to non-CL logic) and even on the site it matters if you have Lazer mode enabled or not.

I implemented it this way because this would make the most sense, with CL plays (only possible on stable) following stable logic and non-CL plays (only possible on Lazer) following Lazer logic

PerformanceCalculator/Simulate/ManiaSimulateCommand.cs
minisbett: I just saw you're accessing the string array. Instead, add `Mod[] mods` to the parameters of `GetAccuracy` [here](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/SimulateCommand.cs#L89) and in all implementations, and then you pass it to the method [here](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/SimulateCommand.cs#L73).

minisbett:
> GetMaxCombo returns the Lazer max combo so it's not a viable option
right. Wonder if that can be PRd to Lazer instead but I'm not sure how much sense that'd make because CL basically only acts as a stable marker on Lazers' mania.

MaxOhn:
`generateHitResults` still assumes perfects and greats to generate the same accuracy value but `GetAccuracy` distinguishes the two by either `305` or `300` values.

MaxOhn:
> `generateHitResults` still assumes perfects and greats to generate the same accuracy value but `GetAccuracy` distinguishes the two by either `305` or `300` values.

The math works the same way so it'll look very similar

```csharp
// Let Perfect=61, Great=60, Good=40, Ok=20, Meh=10, Miss=0. The total should be this.
int targetTotal = (int)Math.Round(accuracy * totalHits * 61);

// Start by assuming every non miss is a meh
// This is how much increase is needed by the rest
int remainingHits = totalHits - countMiss;
int delta = targetTotal - 10 * remainingHits;

// Each perfect increases total by 51 (perfect - meh = 51)
int perfects = Math.Min(delta / 51, remainingHits);
delta -= perfects * 51;
remainingHits -= perfects;

// Each great increases total by 50 (great - meh = 50)
int greats = Math.Min(delta / 50, remainingHits);
delta -= greats * 50;
remainingHits -= greats;

// Each good increases total by 30 (good - meh = 30).
countGood = Math.Min(delta / 30, remainingHits);
delta -= countGood.Value * 30;
remainingHits -= countGood.Value;

// Each ok increases total by 10 (ok - meh = 10).
int oks = delta / 10;
remainingHits -= oks;

// Everything else is a meh, as initially assumed.
countMeh = remainingHits;
```
Just need to check whether stable or lazer accuracy is required and chose the formula accordingly.

MrHeliX:
> > `generateHitResults` still assumes perfects and greats to generate the same accuracy value but `GetAccuracy` distinguishes the two by either `305` or `300` values.
> 
> The math works the same way so it'll look very similar
> 
> ```cs
> // Let Perfect=61, Great=60, Good=40, Ok=20, Meh=10, Miss=0. The total should be this.
> int targetTotal = (int)Math.Round(accuracy * totalHits * 61);
> 
> // Start by assuming every non miss is a meh
> // This is how much increase is needed by the rest
> int remainingHits = totalHits - countMiss;
> int delta = targetTotal - 10 * remainingHits;
> 
> // Each perfect increases total by 51 (perfect - meh = 51)
> int perfects = Math.Min(delta / 51, remainingHits);
> delta -= perfects * 51;
> remainingHits -= perfects;
> 
> // Each great increases total by 50 (great - meh = 50)
> int greats = Math.Min(delta / 50, remainingHits);
> delta -= greats * 50;
> remainingHits -= greats;
> 
> // Each good increases total by 30 (good - meh = 30).
> countGood = Math.Min(delta / 30, remainingHits);
> delta -= countGood.Value * 30;
> remainingHits -= countGood.Value;
> 
> // Each ok increases total by 10 (ok - meh = 10).
> int oks = delta / 10;
> remainingHits -= oks;
> 
> // Everything else is a meh, as initially assumed.
> countMeh = remainingHits;
> ```
> 
> Just need to check whether stable or lazer accuracy is required and chose the formula accordingly.

I refactored this method so it handles both types of logic similarly to the rest of the class

MaxOhn:
Found two remaining flaws on edge cases with my last snippet as mentioned above, the rest looks fine to me 👍

MrHeliX:
Thanks, should be all good now

stanriders:
You've forgot to update hitresult generation for GUI

MrHeliX:
> You've forgot to update hitresult generation for GUI

Oops, updated that as well

stanriders (approved):

PR #256, opened 2025-04-03
Replace manual attribute table building with an `AttributesTable` component

stanriders:
`SimulateScreen` became insanely bloated over time, so it's time to start cutting it into components. This is the first one of hopefully many.

As a side-effect of this change you can now copy attribute values from the table and the table itself is more condensed which makes it easier to see all of it without scrolling

image


tsunyoku:
Unrelated to these changes but we really should make the tick & tail miss boxes less squished at some point.

tsunyoku (approved):

stanriders:
> Unrelated to these changes but we really should make the tick & tail miss boxes less squished at some point.

yea, I'm planning to address that too

PR #259, opened 2025-04-14
Update PerformanceCalculator's documentation based on the app's CLI help

NathanRazaf:


bdach:
As I said elsewhere on discord I'd prefer these were just deleted and the user was directed to use `--help` from CLI directly. Keeping these in sync is a waste of collective time.

PR #263, opened 2025-05-13
Use `LegacyTotalScore` if set when parsing `ScoreInfo`

tsunyoku:
Noticed when trying https://github.com/ppy/osu/pull/33066. Currently, this is going down the path of estimating legacy score for stable scores and was causing differing values.


stanriders:
Oh this is going to be a pain to support isn't it

stanriders (approved):

PR #269, opened 2025-06-22
Allow legacy score id for "populate from score"

Givikap120:
It changes score input field to text so you can enter scores from legacy db


stanriders (changes requested):
This is a very dirty implementation. You're lacking any kind of validation now, as well as making score id fields not actually contain score id

How useful is this? Legacy score links should not be exposed anywhere anymore I'm pretty sure, why would we need to support them?

wulpine:
> If you have legacy link (happens when you take score from lb without lazer mode)

are you by any chance using osuplus? It replaces new links with the legacy ones
> you need to go through painful process of extracting new id from json response

might help you save some time—the "Download Replay" button always has the new score ID

Givikap120:
Added the validation

PR #270, opened 2025-06-22
Use async calculations in SimulateScreen

Givikap120:
This should decrease freezes when changing mods.

### DIFF
--- a/PerformanceCalculatorGUI/Screens/SimulateScreen.cs
+++ b/PerformanceCalculatorGUI/Screens/SimulateScreen.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using osu.Framework;
 using osu.Framework.Allocation;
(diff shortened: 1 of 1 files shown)


stanriders:
I'm not sure how much I'd like to riddle the code with `task.run`'s for minor gains - I'd rather see if we can leverage update thread for difficulty calculation instead. @tsunyoku what do you think?

stanriders:
https://github.com/user-attachments/assets/89e1cdab-5ebf-41f1-b30f-e7817e2cd16c

I don't see the vision sorry - this try was done on a fresh pc boot using hdd so it can be a lot faster on a better setup. Definitely not worth doing adhoc fixes like that for a dev tool, but if you want diffcalc to run without any hitches you should look into refactoring it into an update thread job.

Givikap120:
> 2025-06-28.21-19-51.mp4 
> I don't see the vision sorry - this try was done on a fresh pc boot using hdd so it can be a lot faster on a better setup. Definitely not worth doing adhoc fixes like that for a dev tool, but if you want diffcalc to run without any hitches you should look into refactoring it into an update thread job.

On my setup there was 2 second freeze after every action. I don't see how using Task.Run is adhoc when it's used in diffcalc in other screens like profile screen.

PR #272, opened 2025-06-30
Add ability to calculate legacy scores with scorev1

Givikap120:


PerformanceCalculatorGUI/Screens/SimulateScreen.cs
```diff
@@ -248,6 +249,13 @@ private void load(OsuColour osuColour)
                                                                     notificationDisplay.Display(new Notification("Incorrect score id"));
                                                                 }
                                                             }
+                                                        },
+                                                        legacyScoreSwitchButton = new LabelledSwitchButton
+                                                        {
+                                                            RelativeSizeAxes = Axes.X,
+                                                            Width = 0.25f,
+                                                            Anchor = Anchor.TopLeft,
+                                                            Label = "Is legacy score?",
```
stanriders, line 258: I think it'd be better if we just assumed that if score field is populated then its a legacy score, and if not it's a lazer score.

PerformanceCalculatorGUI/Screens/SimulateScreen.cs
```diff
@@ -869,6 +883,26 @@ private void updateAccuracyParams(bool useFullScoreData)
             }
         }
 
+        private void updateScoreTextBox()
+        {
+            if (ruleset.Value.ShortName == "osu" && legacyScoreSwitchButton.Current.Value)
```
stanriders, line 888: Maybe show score textbox if CL is enabled instead, legacy scores can't be without CL

stanriders (changes requested):

PerformanceCalculatorGUI/Screens/SimulateScreen.cs
stanriders: https://github.com/ppy/osu-tools/pull/272#discussion_r2174863800

PerformanceCalculatorGUI/Screens/SimulateScreen.cs
Givikap120: I think it's too implicit and unintuitive

PR #308, opened 2026-05-18
Change toolbar buttons tooltips to be more visible

stanriders:
Before:
image
After:
image

Hover container in the `PerformanceCalculatorSceneManager` never really worked which made the buttons tooltip text annoying to read. This changes it to be based on normal tooltips logic instead
`ScreenSelectionButton` is now a slimmed down copy of the `ToolbarButton` now

### DIFF
--- a/PerformanceCalculatorGUI/Components/ScreenSelectionButton.cs
+++ b/PerformanceCalculatorGUI/Components/ScreenSelectionButton.cs
@@ -25,9 +25,9 @@ namespace PerformanceCalculatorGUI.Components
         private readonly string title;
         private readonly GlobalAction? hotkey;
 
-        public const float PADDING = 3;
+        private const float padding = 3;
 
-        protected Box HoverBackground;
+        private Box hoverBackground;
         private readonly Box flashBackground;
 
         public ScreenSelectionButton(string title, IconUsage? icon = null, GlobalAction? hotkey = null)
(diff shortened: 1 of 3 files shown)


PerformanceCalculatorGUI/Components/ScreenSelectionButton.cs
```diff
@@ -1,21 +1,221 @@
...
+                        Children = new Drawable[]
+                        {
+                            text = new OsuSpriteText
+                            {
+                                Anchor = Anchor.TopLeft,
+                                Origin = Anchor.TopLeft,
+                                Shadow = true,
+                                Font = OsuFont.GetFont(size: 18, weight: FontWeight.Bold)
+                            },
+                            subTooltipFlow = new FillFlowContainer
```
peppy, line 181: is this hotkey stuff actually used in this project?

stanriders: Settings can be opened using ctrl+o

peppy: Feels like a lot of faff to be inlining for one usage most people already know, dunno.

stanriders: True, but I do want to add screen switching hotkeys at some point too and those definitely will need a tooltip

peppy (approved):

PR #312, opened 2026-09-02
Add favourites dropdown to `ProfileScreen`

stanriders:
image


tsunyoku (approved):
If we keep adding more of these then it may make more sense to add an actual database (sqlite, Realm) rather than adding new JSON files but its fine for now

stanriders:
I'd rather not use binary formats for stuff like collections/favourites, the point of using jsons is for ease of sharing and manual editing

tsunyoku:
For collections that's a fair challenge, not sure about this though.

PR #314, opened 2026-09-07
Use dynamic api version header

stanriders:
Same as https://github.com/ppy/osu/blob/master/osu.Game/Online/API/APIAccess.cs#L91-L92


tsunyoku:
Have you checked the list of breaking changes between the old API version and the latest to make sure this won't have any knock-on effects?

stanriders:
We're using game's DTOs already so if anything having the version pinned is more incorrect. Other than that I've checked everything and it still works as expected, the only thing I've encountered was https://github.com/ppy/osu/pull/38754 which isnt tools fault to begin with

tsunyoku (approved):

PR #315, opened 2026-09-10
Fix taiko hit result generation when only misses are provided

minisbett:
I ran across this issue when trying to simulate a taiko score with just missing combo and misses, but no goods or accuracy specified.

Example: `dotnet run -- simulate taiko 132889 --combo 758 --misses 11`

This causes performance calculation to throw, due to a `NaN` caused by a greats count higher (958) than possible and a goods count in the negative (-22).

The issue here is that it tries to calculate the greats/goods count from a `targetTotal`, which is calculated via `accuracy` and `totalResultCount`:
```cs
int hitCount = totalResultCount - countMiss;
int targetTotal = (int)Math.Round(accuracy * hitCount * 2);

countGreat = targetTotal - (totalResultCount - countMiss);
countGood = totalResultCount - countGreat - countMiss;
```
If an `accuracy` of `1` is passed, along with a `countMiss` of `11`, it assumes the accuracy includes the misses. This is incorrect behavior, and instead it should be `accuracy * (totalResultCount - countMiss) * 2`.

The idea that `accuracy` refers to the hit portion of the hit results is also seen in the [`generateHitResults` method of the osu-ruleset](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Simulate/OsuSimulateCommand.cs#L74).

### DIFF
--- a/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
+++ b/PerformanceCalculator/Simulate/CatchSimulateCommand.cs
@@ -51,9 +51,9 @@ namespace PerformanceCalculator.Simulate
             // Max value minus whatever misses are left. Negative if impossible missCount
             int countFruits = maxFruits - (countMiss - (maxDroplets - countDroplets));
 
-            // Either given or the max amount of hit objects with respect to accuracy minus the already calculated fruits and drops.
-            // Negative if accuracy not feasable with missCount.
-            int countTinyDroplets = countMeh ?? (int)Math.Round(accuracy * (maxCombo + maxTinyDroplets)) - countFruits - countDroplets;
+            // Either given or the max amount of "hit" hit objects with respect to accuracy minus the already calculated fruits and drops.
+            int relevantResultCount = maxCombo + maxTinyDroplets - countMiss;
+            int countTinyDroplets = countMeh ?? (int)Math.Round(accuracy * relevantResultCount) - countFruits - countDroplets;
 
             // Whatever droplets are left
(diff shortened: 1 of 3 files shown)


stanriders:
Would you mind doing it in gui as well?

minisbett:
Sure. I've also came across issues with catch, where adding misses turns another hit result negative, but if anything I'll address that in another PR.

stanriders:
Just do it in this PR if it's something simple

minisbett:
The issue I have with catch is apparently intended.

https://github.com/minisbett/osu-tools/blob/b471efb6a50e027d7d2c7134a7b439cb60558762/PerformanceCalculator/Simulate/CatchSimulateCommand.cs#L55

Here, an accuracy of `1` + misses is intentionally invalid, while in the other modes the accuracy refers to the hit portion. So dunno if this should be changed for consistent behavior, since if you just want the score + 2 misses, you will need to calculate the accuracy for that yourself. I have a fix for that, but I'd say that's your call. But I would say it would make more sense to assume accuracy refers to the hit portion for CLI purposes, since when not specifying it always default to 1.

stanriders:
I think we should have all rulesets behave the same, so it would make more sense to change catch generation in this case

minisbett:
Done. All four modes now consistently have accuracy refer to the hit portion, allowing to specify misses while the accuracy is 1 in perfcalcgui, or not specified as an argument at all in perfcalc (where it defaults to 1).

stanriders (approved):

